PR #18, opened 2023-04-12
fix(quora): Some times get repetitive message

s524797336:
Some times get repetitive message
eg: Helllloooowww


xtekky:
I will have to test it, would you be able to give more details ?

xtekky:
closing for now until this issue gets addressed again

PR #89, opened 2023-04-23
Fix for Poe account creation

sudouser777:
- added new class `Poe` this will be used as an entry point
-  added selenium for signup/login into poe.com
-  added code to save cookie to a file, this will be used from next time onwards, if expires it will try to register again
-  added fake-useragent module, we can use this instead of static string

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,16 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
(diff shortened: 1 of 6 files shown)


sudouser777:
@xtekky please check the changes and test once, I will add the bot creation logic in couple of days

xtekky (changes requested):
remove `|` in typing as this operator is not supported in previous python versions

xtekky:
works fine, will add it now ; )

PR #103, opened 2023-04-24
updated structure of the "you" module

prtolem:
created a new file `you.py` where is the main code located. (the code was restored to normal and redone the work with `chat`)
updated the file `__init__.py` where now there is only import from the main file.
updated the example in the readme


xtekky:
what is the utility of this apart of refractoring it so it does not look like other models

you = Completion() ??

PR #191, opened 2023-04-27
Update README.md

DanielShemesh:
fixed table


adlai:
Something was wrong on the Internet, and you fixed it for free? Must be popular with party.

PR #200, opened 2023-04-27
Reformat code using PyCharm

AymaneHrouch:
Please review and merge so I can push new code, the code was automatically formatted using PyCharm with no intervention from my part.


sudouser777:
all people may not use pycharm, it's better to use [black](https://pypi.org/project/black/) I think

AymaneHrouch:
@sudouser777 I just wanted to format it as a start because it's really messy now. If black sounds like a good option we can use it too.
@xtekky any thoughts?

xtekky:
it has conflics, please format the newer file

AymaneHrouch:
@xtekky done

forefront/__init__.py
```diff
@@ -1,51 +1,59 @@
...
 from requests import post
-from json import loads
+from tls_client import Session
+
+from forefront.mail import Mail
+from forefront.typing import ForeFrontResponse
 
 
 class Account:
+<<<<<<< HEAD
```
53jk1, line 14: Conflicts

forefront/__init__.py
```diff
@@ -1,51 +1,59 @@
...
     @staticmethod
     def create(proxy = None, logging = False):
         
+=======
```
53jk1, line 18: Conflicts

forefront/__init__.py
```diff
@@ -1,51 +1,59 @@
...
+    def create(proxy=None, logging=False):
+
+>>>>>>> e627601 (Reformat code using PyCharm)
```
53jk1, line 21: Conflicts

53jk1:
@AymaneHrouch I would like to suggest considering the implementation of a universal linter to ensure consistent code formatting across different development environments. Since different developers may have different configurations for their PyCharm IDE, having a shared linter configuration will ensure that everyone follows the same style guidelines regardless of their individual IDE settings.

EZSNoVa:
Black is a good option, since it it available in most IDE's or text editors

PR #234, opened 2023-04-28
Deleted Singularity Directory

naa7:
There is no need for this directory.


editor-syntax:
How come it's not needed?

PR #240, opened 2023-04-28
openaihosted ref

ezerinz:
refactor openaihosted, do something to convert escape sequence code, added example in readme


xtekky:
Hi, openaihosted was also refractored but I like ur code ; ), you can submit a new pr if you wish

ezerinz:
okay

PR #254, opened 2023-04-28
Print the mail messages only when logging

0dminnimda:
It's annoying ^-^

### DIFF
--- a/forefront/__init__.py
+++ b/forefront/__init__.py
@@ -60,7 +60,7 @@ class Account:
         while True:
             sleep(1)
             for _ in mail.fetch_inbox():
-                print(mail.get_message_content(_["id"]))
+                if logging: print(mail.get_message_content(_["id"]))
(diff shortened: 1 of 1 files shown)


Ranu2001:
import logging

logging.basicConfig(level=logging.DEBUG,
                    format='%(asctime)s %(levelname)s %(message)s')

logger = logging.getLogger(__name__)

# Set the log message level for printing mail messages
mail_level = logging.INFO

# Example mail message
mail_message = "New mail received!"

# Log a mail message only if the log message level is set to mail_level or higher
if logger.isEnabledFor(mail_level):
    logger.log(mail_level, mail_message)

In this example, we set the log message level for printing mail messages to INFO, but you can modify this to your specific needs. The is enabled for method checks if the logger is enabled for the specified log level, and the log method prints the mail message only if the logger is enabled for the specified log level or higher.

By using this conditional statement, you can ensure that mail messages are printed only when the appropriate logging level is set, improving the efficiency of your code and reducing unnecessary print statements.

xtekky:
mb

PR #280, opened 2023-04-29
Update README.md

mishalhossin:


mishalhossin:
added `from gpt4free.quora import Poe`
instead of `from quora import Poe`

PR #285, opened 2023-04-29
forefront: update account creation logic

Akash98Sky:
- [x] Update account creation and verification logic, as per the latest forefont.ai APIs.
- [x] Use the official mail.tm API library [pymailtm](https://github.com/CarloDePieri/pymailtm).


Vlad202:
is it possible to make code faster? account reg works very slow. Am i need to register account per everty request?

Akash98Sky:
Is it possible to make code faster?
- For account creation, all these steps are necessary and it is a one-time process. It would take around 10-15 secs, which is fine I guess.

Am I need to register an account per every request?
- No. On successful account creation email address and token would be dumped in a file named accounts.txt, these tokens can be used to make future requests.
Not sure how long these tokens are valid. But yeah we can look into it about how to reuse an account.

PR #304, opened 2023-04-30
backup options to Emailnator

taiyi747:
Emailnator seems to be a bit unstable, and there may be emails already in the original mailbox. 

This code is completely modeled after the original Mail.py.


taiyi747:
**Emailnator has some pre-existing emails in the mailbox, which may affect the program's judgment. This new code uses an edu mailbox, which is relatively stable.**

PR #438, opened 2023-05-04
gpt4free typescript version

xiangsx:
I have written a TypeScript version of GPT4Free based on your project and I hope to join the GPT4Free family with my contribution.

I have only implemented model for you.com, but I hope to join your project in order to attract more people to join and integrate more models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,13 @@ Just API's from some language model sites.
       <td><a href="https://github.com/xtekky/gpt4free/pulls"><img alt="Pull Requests" src="https://img.shields.io/github/issues-pr/xtekky/gpt4free?style=flat-square&labelColor=343b41"/></a></td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


xtekky:
hi @xiangsx, this is very interesting, as I have been struggling to find a tls-client that accepts streaming in python, and it works in `.ts`, I might be switching language or create python bindings ; ), very cool.

xiangsx:
> hi @xiangsx, this is very interesting, as I have been struggling to find a tls-client that accepts streaming in python, and it works in `.ts`, I might be switching language or create python bindings ; ), very cool.

I am really glad that my idea aligns with yours, but unfortunately tls-client does not support streams on the node side. I wait for the response data and convert it into stream data afterwards. I'll keep looking for other solutions.

xtekky:
Oh, I didnt see this haha, thanks

PR #509, opened 2023-05-08
Implemented the ability to save forefront cookies

MIDORIBIN:
Added the ability to save cookies in the create function.
Default is disabled.

Refactoring is being done as it has become redundant code.
* split funcion
* delete account.txt(unused)

Similar to the following pull request, but since it was removed, I am suggesting it again.
https://github.com/xtekky/gpt4free/pull/386

Time taken to process `Account.create`

Cookies disabled(default): 9.393363237380981
Cookies enabled: 0.5772099494934082

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,9 +14,6 @@ venv/
 conversations.pkl
 *.pkl
 
-# Ignore accounts created by api's
-accounts.txt
-
 .idea/
 
 **/__pycache__/
@@ -30,3 +27,5 @@ cookie.json
 *.pyc
 
 dist/
+
+gpt4free/forefront/cookie.txt
--- a/gpt4free/forefront/__init__.py
+++ b/gpt4free/forefront/__init__.py
@@ -1,4 +1,5 @@
 import hashlib
+import os
 from base64 import b64encode
 from json import loads
 from re import findall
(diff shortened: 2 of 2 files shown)


xiangsx:
Have you tested?
forefront jwttoken will expire in 1 minutes and limit gpt4 5times/day

so you can keep jwttoken alive by send update token every 50s
when up to 5times gpt4 ask, sign up again or change to gpt-3.5-turbo.
I think this is the best solution.

you can try, follow this commit.
https://github.com/xiangsx/gpt4free-ts/commit/67008b75a2f5f6e2132c94e530de88396cd7a6a8#diff-69760a5c7ac0b0b2df74423d2ced74cbb5788e6922148d15a8d7b18e4bad86a0R215

MIDORIBIN:
I'm testing if it works.

> forefront jwttoken will expire in 1 minutes and limit gpt4 5times/day

I know both, but don't see it as a problem.
**The use of cookies is disabled by default.**
This pull request **does not force the use** of cookies. It only **allows** them to be used.

If someone wants to use gpt-4 for more than a minute or 5 times a day, they can just use the default. (no cookies).
If someone wants to use only gpt-3.5 without creating an account, they can use cookies.

MIDORIBIN:
> you can try, follow this commit.

this implementation is great.

However, it is not related to this pull request, so I will implement it in another pull request.

MIDORIBIN:
@xtekky 
Is there a problem?

PR #532, opened 2023-05-10
Fix Chinese garbled characters

with1874:
When getting a reply in Chinese, it will be garbled. Use utf-8 decoding to solve it.


taiyi747:
This problem seems to have occurred in #241, and I don't understand this utf8 encoding and decoding...

PR #542, opened 2023-05-12
repair, add module: Crypto

nongfu9woo:
Fixing the error when using forefront: ModuleNotFoundError: No module named 'Crypto'.


ilyasdc:
The module already exists, it's in `pycryptodome`, you're probably running another python than the one in your virtual environment.

PR #559, opened 2023-05-16
fix query_forefront call argument bug

itegel:
Only a small bug, please see the diffs ...


phamxtien:
With below code I get return: **Expecting value: line 1 column 1 (char 0)**
```
from gpt4free import forefront
    try:
        account_data = forefront.Account.create(logging=False)
        ans = forefront.Completion.create(account_data=account_data, prompt='hello', model='gpt-4')
        print(ans.text)
    except Exception as e:
        print(str(e))
```

majid-farsijani:
[Mazare.madar.ir](url)

PR #568, opened 2023-05-19
add oraai

hpsj:


xtekky:
ora.sh got taken down already, i appreciate your efforts though

PR #598, opened 2023-05-23
Fixed error in gpt4free/usesless/__init__.py [IndexError: list index out of range]

enganese:
### **Fixed issue**:
```
message_id = mail_client.message_list()[0]["id"]  # <--- Error raised here
IndexError: list index out of range
```

### **Added**:
```
...
while True:
        time.sleep(5)
        messages = mail_client.message_list()
        
        # Check if method `message_list()` didn't return None or empty list.
        if not messages or len(messages) == 0:
            # If it returned None or empty list just 'continue'.
            continue
    
...
```

### **Now, email confirmation looks like this**:
```
...
while True:
        time.sleep(5)
        messages = mail_client.message_list()
        
        # Check if method `message_list()` didn't return None or empty list.
        if not messages or len(messages) == 0:
            # If it returned None or empty list just 'continue'.
            continue
        
        message_id = messages[0]["id"]  # <--- Error was here
        message = mail_client.message(message_id)
        verification_url = re.findall(
            r"http:\/\/ai\.usesless\.com\/api\/cms\/auth\/email-confirmation\?confirmation=\w.+\w\w",
            message["text"],
        )[0]
        if verification_url:
            break
...
```

### **Confirmation**:
**_No need to test the snippet. Code was tested just by running and it prevents IndexError. However, the confirmation message is not being received on temporary mail (not my problem, but we will try to determine the error and find a new solution)_**

### DIFF
--- a/gpt4free/usesless/__init__.py
+++ b/gpt4free/usesless/__init__.py
@@ -56,7 +56,7 @@ class Account:
             messages = mail_client.message_list()
             if not messages or len(messages) == 0:
                 continue
-            message_id = mail_client.message_list()[0]["id"]
+            message_id = messages[0]["id"]
             message = mail_client.message(message_id)
             verification_url = re.findall(
                 r"http:\/\/ai\.usesless\.com\/api\/cms\/auth\/email-confirmation\?confirmation=\w.+\w\w",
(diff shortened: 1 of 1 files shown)


enganese (commented):
_My first commit / change doesn't fit the email confirmation approach of @ezerinz, so I refactored it again._ 

**Now, it perfectly fits to its initial logic.**

PR #602, opened 2023-05-25
Remove forced Streamlit version

Josh-XT:
Remove forced Streamlit version. You're breaking other software that uses GPT4Free and Streamlit for their own apps by forcing this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ pypasser
 names
 colorama
 curl_cffi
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

MuhammadMannan (approved):

xtekky:
fixed, also couldnt find the file, i think i am getting crazy

PR #610, opened 2023-05-26
Added support for AiColors

ggindinson:
### Link to AiColors website - https://aicolors.co

This AI generates color palletes from query. I wrote clean, extendable code, used dataclasses for better responses and did my best. You can find the code in the gpt4free folder.

Please, merge it! It can be very useful for artists and designers, or just creative people =)


enganese:
@xtekky, he is my friend. 

**We tested the code whole night, I can guarantee the operationality. **
Also, `aicolors` has no limit for usage (even without account). 

_FYI, if you send 1,000,000 requests and 100 at a time it will handle them all and response with status_code 200._

_**Average response time is 4-5 seconds(afaik from @ggindinson)**_

PR #620, opened 2023-05-28
Update __init__.py (refactored import)

enganese:
### **Before**:
`from typings import AiColorsResponse`

### **After**:
`from .typings import AiColorsResponse`

### **Comment**:
`Now, it perfectly works!`


Kiri2002:
you r so great!! thx! it works!!

PR #659, opened 2023-06-10
g4f.ai APP

Ron-Caster:
An app for G4F. It makes accessibility easy right?


Ron-Caster:
Previously I made the web app for using xtekkys gptforfree. It wasn't working. So after a long time I realised that a new website is realesd for G4F. For ease of access i made it to an app. I use it personally. And I used to send an issue whenever I found something.

xtekky:
would you be able to include more info ?  I think best solution is to make another repository with source code, and I can add it to the shared projects list, but I like the idea !!

Issue #660, opened 2023-06-11
gpt4free useless: IndexError: list index out of range


maksc1:
**Bug description**
Telegram bot using gpt4free not working
main.py:
```import telebot
from gpt4free import usesless

bot = telebot.TeleBot('my_token')

@bot.message_handler(commands=['start'])
def send_welcome(message):
    bot.reply_to(message, "ChatGPT unlimited and free but without memory")

@bot.message_handler()
def test(message):
  prompt = ""
  req = usesless.Completion.create(prompt=prompt)
  prompt = message.text
  bot.reply_to(message, req["text"])

if __name__ == "__main__":
  bot.polling()
```
Error:
```
Traceback (most recent call last): File "main.py", line 20, in <module> bot.polling()

File "/home/runner/Test/venv/lib/python3.1

0/site-packages/telebot/__init__.py", line 1 043, in polling self.__threaded_polling (non_stop=non_sto p, interval=interval, timeout=timeout, long_

polling_timeout-long_polling_timeout,

File "/home/runner/Test/venv/lib/python3.1

0/site-packages/telebot/__init__.py", line 1 118, in __threaded_polling

raise e

File "/home/runner/Test/venv/lib/python3.1 0/site-packages/telebot/__init__.py", line 1 074, in threaded_polling

self.worker_pool.raise_exceptions() File "/home/runner/Test/venv/lib/python3.1 0/site-packages/telebot/util.py", line 147, in raise_exceptions

raise self.exception_info File "/home/runner/Test/venv/lib/python3.1 0/site-packages/telebot/util.py", line 90, i n run

task(*args, **kwargs) File "/home/runner/Test/venv/lib/python3.1 0/site-packages/telebot/__init__.py", line 6 770, in _run_middlewares_and_handler result = handler['function'](message)

File "main.py", line 15, in test

req = usesless.Completion.create(prompt=prompt) File "/home/runner/Test/venv/lib/python3.1 0/site-packages/gpt4free/usesless/__init__.py", line 46, in create

response = Completion.__response_to_json (content) File "/home/runner/Test/venv/lib/python3.10/site-packages/gpt4free/usesless/__init__.py", line 53, in __response_to_json split_text = text.rsplit("\n", 1)[1]

IndexError: list index out of range
```

**Environement**
- python version 3.10
- server location Poland

**Additional context**
If you need more information to help me, please let me know.

LopeKinz:
It seems like you are missing some arguments in your code. here is a example usage : 
[...]

PR #664, opened 2023-06-13
Fix for issue #660

LopeKinz:
#660 Fix

### DIFF
--- a/gpt4free/usesless/__init__.py
+++ b/gpt4free/usesless/__init__.py
@@ -144,10 +144,15 @@ class Completion:
         response = Completion.__response_to_json(content)
         return response
 
+
     @classmethod
     def __response_to_json(cls, text) -> str:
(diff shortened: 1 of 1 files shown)


LopeKinz:
The error in the module is occurring because the response content from the API call is not in the expected format. To fix the error, you need to update the `__response_to_json` method in the `Completion` class. Here's the updated code for that method:

```python
@classmethod
def __response_to_json(cls, text) -> str:
    text = str(text.decode("utf-8"))

    split_text = text.rsplit("\n", 1)
    if len(split_text) > 1:
        to_json = json.loads(split_text[1])
        return to_json
    else:
        return None
```

With this updated code, the method checks if `split_text` has more than one element before accessing the second element. If there's only one element, it returns `None` instead. This change will prevent the `IndexError` when the response content doesn't have the expected format.

Make sure to update the module with this code, and the error should be resolved.

PR #705, opened 2023-06-29
Fix: change subDomain of LockChat

tuanle96:
Recently, I changed to new domain and this provider worked.


killporter:
i get an empty response

xtekky:
looks like this should work

PR #706, opened 2023-06-30
Update:Make streaming responses more like the OpenAI API.

kncxstudio:
Update:Make streaming responses more like the OpenAI API.

### DIFF
--- a/interference/app.py
+++ b/interference/app.py
@@ -1,27 +1,30 @@
-import os
 import time
 import json
 import random
+import time
 
-from g4f import Model, ChatCompletion, Provider
-from flask import Flask, request, Response
+from flask import Flask, request
 from flask_cors import CORS
+
+from g4f import ChatCompletion, Provider
 
 app = Flask(__name__)
 CORS(app)
 
+
 @app.route("/chat/completions", methods=['POST'])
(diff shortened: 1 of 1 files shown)


nacho00112:
Why you didn't used the `string` standard library to generate the `completion_id`?

iloveyumyumshrimpnoodles:
> Why you didn't used the `string` standard library to generate the `completion_id`?

Why import another entire library, just for a few characters?

interference/app.py
```diff
@@ -1,27 +1,30 @@
...
-    response = ChatCompletion.create(model=model, stream=streaming,
+
+    response = ChatCompletion.create(provider=Provider.You, model=model, stream=streaming,
                                      messages=messages)
-    
+
     if not streaming:
         while 'curl_cffi.requests.errors.RequestsError' in response:
-            response = ChatCompletion.create(model=model, stream=streaming,
+            response = ChatCompletion.create(provider=Provider.You, model=model, stream=streaming,
```
w3bdesign, line 27: Why hardcode model here?

w3bdesign (changes requested):
Why hardcode model?

PR #724, opened 2023-07-04
fix you.com provider

adi6409:
Moved the params variable to be before the header, and replaced the header referer with params to make it work.


gulldan:
seems not work=(

phamxtien:
It returns nothing

xtekky:
fix incomming soon

PR #726, opened 2023-07-05
added some more providers

najam-tariq:
1. AItianhu
2. EasyChat
3. Acytoo
4. DFEHub


xtekky:
good job !

phamxtien:
1. g4f.Provider.Providers.Acytoo provider returns <Response [429]>
2. g4f.Provider.Providers.DFEHub provider return Error Occurred::405

najam-tariq:
@phamxtien 
1. Acytoo is not working atm — technical issue on the Provider's side
2. Fixed DFEHub temporarily by implementing support for rate-limit, but I think @xtekky would be able to come up with a better solution (probably bypass rate-limit?)

phamxtien:
EasyChat got error: 
Error Occurred::403

Issue #153, opened 2023-04-26
phind.com takedown request


xtekky:
Unfortunately I will have to take down code for phind.com soon,

<img width="780" alt="image" src="https://user-images.githubusercontent.com/98614666/234576741-caa88f51-ff7b-4e85-911b-d2a19c890793.png">

will be replaced by an alternative gpt-4 i am working on at the moment

9xN:
damn

cometbus:
Is there a work around? is this broken for now?

xtekky:
its still working atm and will prolly be revived by another repo, watch out I am not the only reverse egineer on github...

9xN:
> its still working atm and will prolly be revived by another repo, watch out I am not the only reverse egineer on github...

i dont believe it is working atm but i might make a fork and make a repo for it

AbdelrhmanNile:
> 
its not working

mishalhossin:
> > its still working atm and will prolly be revived by another repo, watch out I am not the only reverse egineer on github...
> 
> i dont believe it is working atm but i might make a fork and make a repo for it

It's working 
![image](https://user-images.githubusercontent.com/91066601/235299786-db8aec2b-52c3-4f44-8dc7-75e02288b2df.png)

Pozaza:
> It's working ![image](https://user-images.githubusercontent.com/91066601/235299786-db8aec2b-52c3-4f44-8dc7-75e02288b2df.png)

can you send me the code?

mishalhossin:
> > It's working ![image](https://user-images.githubusercontent.com/91066601/235299786-db8aec2b-52c3-4f44-8dc7-75e02288b2df.png)
> 
> can you send me the code?

not working now

0dminnimda:
> > > It's working ![image](https://user-images.githubusercontent.com/91066601/235299786-db8aec2b-52c3-4f44-8dc7-75e02288b2df.png)
> > 
> > 
> > can you send me the code?
> 
> not working now

because of cloudflare?

PR #744, opened 2023-07-12
Added Phind to Providers

AlephZero255:
Added Phind to Providers. May not always work consistently, but most of the time it's fine.


bagusindrayana:
https://github.com/xtekky/gpt4free/issues/153
i thought we can't use Phind anymore (?)

AlephZero255:
> #153 i thought we can't use Phind anymore (?)

Yeah, I guess you're right(

PR #761, opened 2023-07-19
Update README.md

notbrianbob:
Changed the print loop to print the entire response in a single string, rather than a \n after every character


MIDORIBIN:
@notbrianbob 
The modification will make it no longer a Stream.
If line breaks are not necessary, the following fix is recommended:

```
for message in response:
    print(message, end='', flush=True)
```

PR #770, opened 2023-07-22
fix import bug when install it in local env.

gasvn:


xtekky:
could you please resolve the conflicts, I will then merge it

PR #771, opened 2023-07-24
Bring back the README banner

nonk123:
Whatever the new banner was, the link to it specified in the README died, so I replaced it with the v1 banner. Better that than nothing, right?


xtekky:
image

i dont see any issue

nonk123:
Weird. That's what I'm seeing:
image
And the link is a 404. Maybe the original link is referring to a private repo only available to you?

I can't actually verify if anyone else is having this issue because no one has reported it, but getting a 404 on a GitHub repo is definitely not a good sign.

nonk123:
Actually, if you do have access to the new image, could you send the original file here? Uploading it to the repo (e.g. into the assets directory) and referring to the file inside the README would be a much better idea than using an external link that isn't verified to work for everybody.

PR #800, opened 2023-08-13
refactor: refactor provider

MIDORIBIN:
@xtekky 

* Refactored the provider:
  * Created an abstract class called `BaseProvider` and modified all providers to inherit from that class.
* Enabled providers Vercel, You, etc.
* Performed type annotations.
* Delete v1
  * It can be retrieved from the git history if needed

If you have any questions due to these significant changes, please feel free to ask.


xtekky:
could you resolve the conflicts

MIDORIBIN:
@xtekky 
Conflicts were resolved.

MIDORIBIN:
@xtekky 
https://github.com/xtekky/gpt4free/blob/d53fc931a3cb059851eae7f4ffc1f515d4091998/g4f/__init__.py#L9

It looks like you are using "or" for the union type.
In python use "typing.Union" or "|" symbol.

ref: https://peps.python.org/pep-0604/

xtekky:
| is not compatible for lower python versions, thus the "or"

xtekky:
This update involves a lot of changes, so I'll have to look into it further before merging

MIDORIBIN:
@xtekky 
> | is not compatible for lower python versions, thus the "or"

I checked pep and could not find "or" as a symbol for multiple types.
Perhaps it is a meaningless symbol. (Python doesn't check for type hints at runtime, so I'm assuming the incorrect notation isn't causing any errors.)

Therefore, if you want to support python 3.9 or lower, you should use Union[hoge, fuga].
My personal opinion is that it is very hard to support older versions, so it is better to support 3.10 or higher.

MIDORIBIN:
@xtekky 
> This update involves a lot of changes, so I'll have to look into it further before merging

ok.
Let me know if there is anything I can do to help.

xtekky:
I'll be using Union[str, str] from now on, was using "or" because | is an or operator, merging soon, I also have to update the guide to contribute as the structure is quite different and I believe more complex

xtekky:
nice update !!

PR #818, opened 2023-08-17
interference/app.py: remove unused imports

0dminnimda:
Also `Model` is no longer accessible from g4f, so it causes errors, since it is not used there is no point importing it from `g4f.models`

### DIFF
--- a/interference/app.py
+++ b/interference/app.py
@@ -3,8 +3,8 @@ import time
 import json
 import random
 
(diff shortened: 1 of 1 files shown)


0dminnimda:
Oh well, i guess my for was a little late
This was already done in f6ef3cb2237d8c336e915ef77ddbe6f37934c4fd

PR #832, opened 2023-08-22
Fixed Equing non-streaming request error and added encoding and temperature settings.

zeng-rr:


Luneye (changes requested):
non-streaming response is already handled somewhere else and should not be handled inside providers create_completion functions. Take a look at Vercel model. Here is how it could get modified to fix the whole:
[Equing.txt](https://github.com/xtekky/gpt4free/files/12408021/Equing.txt)

Luneye (approved):

johnd0e:
But it still does not work with stream=True:
```
  File "D:\AI\gpt4free\stream.py", line 13, in <module>
    for message in response:
  File "D:\AI\gpt4free\g4f\Provider\Equing.py", line 63, in create_completion
    line_json = json.loads(line.decode('utf-8').split('data: ')[1])
IndexError: list index out of range
```

johnd0e:
@xtekky [Equing.txt](https://github.com/xtekky/gpt4free/files/12408021/Equing.txt) from @Luneye works for me.

PR #833, opened 2023-08-22
Fix bing provider

hlohaus:


phamxtien:
It returns error:
AsyncProvider.create_completion() got multiple values for argument 'cls'

hlohaus:
This PR hast a fix: https://github.com/xtekky/gpt4free/pull/844/files#diff-2e095978f4420600f8dd5df9c124f4e38189ef0efcfd80694a32f25e16cf1b96L80

phamxtien:
After update the code it returns:
message='Attempt to decode JSON with unexpected mimetype: text/html', url=URL('https://sydney.bing.com/sydney/DeleteSingleConversation')

hlohaus:
Everytime when you use Bing? Or only once? I can use Bing without Problems. You have to Pass the captcha Challenge in the Browser. The Captcha comes when a new Chat are created.

phamxtien:
> Everytime when you use Bing? Or only once? I can use Bing without Problems. You have to Pass the captcha Challenge in the Browser. The Captcha comes when a new Chat are created.

On first time, but get error :(
I'm still trying

PR #827, opened 2023-08-20
update header & response easychat

bagusindrayana:
fix forbidden error
```bash
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://chat4.fastgpt.me/api/openai/v1/chat/completions
```
add one server `https://gxos1h1ddt.fastgpt.me`
update response check (status, length, etc) and exception

PR #838, opened 2023-08-24
fix easychat stream : remove whitspace from data: and raise exception

bagusindrayana:
based problem in https://github.com/xtekky/gpt4free/pull/827#issuecomment-1690770477


johnd0e:
Not related to your fixes, but `Error 403 from server`

bagusindrayana:
> Not related to your fixes, but `Error 403 from server`

Yeah they add some captcha, good for them, they aware that they service getting scraped

PR #847, opened 2023-08-26
Hotfix: Dont't call get_cookies in global space

hlohaus:

PR #853, opened 2023-08-28
remove cookies from global space (#847)

ostix360:
The get_cookies can make some error so remove it from execution if the provider is not used
retake from #847


ostix360:
#855 seem to fix better this issue

PR #854, opened 2023-08-28
Add GoogleBard to requirements

Josh-XT:
Adding GoogleBard to requirements resolves errors due to it not being installed.


Josh-XT:
Looks like the bard module isn't being used, I'll keep digging on the issue.

PR #868, opened 2023-09-01
Fix: Python versions below 3.9 cannot use `g4f`

Lin-jun-xiang:
In versions of Python below 3.9,, the `g4f` cannot use correctly. For example (python3.8):

```python
import g4f

# Set with provider
    model="gpt-3.5-turbo",
    provider=g4f.Provider.Wewordle,
    messages=[{"role": "user", "content": "Hello world"}],
)
print(response)
```

**Result before fix:**
```
  File "d:/Junxiang/Python/gpt4free/test.py", line 1, in <module>
    import g4f
  File "d:\Junxiang\Python\gpt4free\g4f\__init__.py", line 1, in <module>
    from .          import models
  File "d:\Junxiang\Python\gpt4free\g4f\models.py", line 2, in <module>
    from .Provider import Bard, BaseProvider, GetGpt, H2o, Liaobots, Vercel, Equing
  File "d:\Junxiang\Python\gpt4free\g4f\Provider\__init__.py", line 1, in <module>
    from .Acytoo        import Acytoo
  File "d:\Junxiang\Python\gpt4free\g4f\Provider\Acytoo.py", line 6, in <module>
    from .base_provider import BaseProvider
  File "d:\Junxiang\Python\gpt4free\g4f\Provider\base_provider.py", line 9, in <module>
  File "d:\Junxiang\Python\gpt4free\g4f\Provider\base_provider.py", line 21, in BaseProvider
    messages: list[dict[str, str]],
TypeError: 'type' object is not subscriptable
```

**Result after fix:**
```
Hi there! How can I assist you today?
```

---

## Descrpition

Use standard **type hints** like `List` and `Dict` instead of `list` and `dict`. The latter will result in an error in **Python versions below 3.9.**

Refer to the following Stack Overflow discussions for more information:
1. https://stackoverflow.com/questions/75202610/typeerror-type-object-is-not-subscriptable-python
2. https://stackoverflow.com/questions/59101121/type-hint-for-a-dict-gives-typeerror-type-object-is-not-subscriptable

### DIFF
--- a/g4f/Provider/AItianhu.py
+++ b/g4f/Provider/AItianhu.py
@@ -2,7 +2,7 @@ import json
 
 import requests
 
-from ..typing import Any, CreateResult
+from ..typing import Any, CreateResult, Dict, List
 from .base_provider import BaseProvider
 
 
@@ -14,7 +14,7 @@ class AItianhu(BaseProvider):
     @staticmethod
     def create_completion(
         model: str,
-        messages: list[dict[str, str]],
+        messages: List[Dict[str, str]],
         stream: bool, **kwargs: Any) -> CreateResult:
         
         base = ""
@@ -44,7 +44,7 @@ class AItianhu(BaseProvider):
     def params(cls):
         params = [
             ("model", "str"),
-            ("messages", "list[dict[str, str]]"),
+            ("messages", "List[Dict[str, str]]"),
             ("stream", "bool"),
             ("temperature", "float"),
             ("top_p", "int"),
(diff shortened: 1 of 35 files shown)


MIDORIBIN:
If you need to support python less than 3.9, it is better to add the following import
The new syntax can be used for python 3.9 and below.

```
from __future__ import annotations
```

Lin-jun-xiang:
You're right. I should use `__future__` better.

PR #920, opened 2023-09-19
Adding embedding support to the interference proxy

chatgpt-tricks:
Adding embedding support to the interference proxy, using the all-mpnet-base-v2 model. 
I know this is already free but it can be handy in situations where you both need chat completions and embeddings.
by the way it's still kind of sketchy


chatgpt-tricks:
just realized I forgot this line:
from transformers import AutoTokenizer

Issue #913, opened 2023-09-17
TypeError: unhashable type: 'Model'


bluusun:
import g4f, asyncio

async def run_async():
  _providers = [
      g4f.Provider.ChatgptAi,
      g4f.Provider.ChatgptLogin,
      g4f.Provider.DeepAi,
      g4f.Provider.Opchatgpts,
      g4f.Provider.Vercel,
      g4f.Provider.Wewordle,
      g4f.Provider.You,
      g4f.Provider.Yqcloud,
  ]
  responses = [
      provider.create_async(
          model=g4f.models.default,
          messages=[{"role": "user", "content": "Hello"}],
      )
      for provider in _providers
  ]
  responses = await asyncio.gather(*responses)
  for idx, provider in enumerate(_providers):
      print(f"{provider.__name__}:", responses[idx])

asyncio.run(run_async())

TypeError: unhashable type: 'Model'

Version 0.0.3.0

PR #924, opened 2023-09-20
Fix async example in readme

hlohaus:
Improve error handling in ChatgptLogin
Cache "snlm0e" in Bard
Fix issue: https://github.com/xtekky/gpt4free/issues/913

### DIFF
--- a/g4f/Provider/base_provider.py
+++ b/g4f/Provider/base_provider.py
@@ -25,6 +25,7 @@ class BaseProvider(ABC):
         
         raise NotImplementedError()
 
+
     @classmethod
     @property
     def params(cls):
@@ -46,6 +47,8 @@ class AsyncProvider(BaseProvider):
         stream: bool = False,
         **kwargs
     ) -> CreateResult:
+        check_running_loop()
+
         yield asyncio.run(cls.create_async(model, messages, **kwargs))
 
     @staticmethod
@@ -67,10 +70,17 @@ class AsyncGeneratorProvider(AsyncProvider):
         stream: bool = True,
         **kwargs
     ) -> CreateResult:
-        loop = asyncio.new_event_loop()
+        check_running_loop()
+
+        # Force use selector event loop on windows
+        loop = asyncio.SelectorEventLoop()
         try:
-            asyncio.set_event_loop(loop)
-            generator = cls.create_async_generator(model, messages, stream=stream, **kwargs)
+            generator = cls.create_async_generator(
+                model,
+                messages,
+                stream=stream,
+                **kwargs
+            )
             gen  = generator.__aiter__()
             while True:
                 try:
@@ -78,9 +88,7 @@ class AsyncGeneratorProvider(AsyncProvider):
                 except StopAsyncIteration:
                     break
         finally:
-            asyncio.set_event_loop(None)
             loop.close()
-
 
     @classmethod
     async def create_async(
@@ -100,6 +108,11 @@ class AsyncGeneratorProvider(AsyncProvider):
     ) -> AsyncGenerator:
         raise NotImplementedError()
 
+# Don't create a new loop in a running loop
+def check_running_loop():
+    if asyncio.events._get_running_loop() is not None:
+        raise RuntimeError(
+            'Use "create_async" instead of "create" function in a async loop.')
 
 _cookies = {}
 
--- a/g4f/Provider/base_provider.py
+++ b/g4f/Provider/base_provider.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from asyncio import SelectorEventLoop
 from abc import ABC, abstractmethod
 
 import browser_cookie3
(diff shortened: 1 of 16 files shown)


g4f/Provider/base_provider.py
```diff
@@ -108,12 +107,17 @@ def create_async_generator(
...
-def check_running_loop():
+
+def create_event_loop():
+    # Don't create a new loop in a running loop
     if asyncio.events._get_running_loop() is not None:
         raise RuntimeError(
             'Use "create_async" instead of "create" function in a async loop.')
 
+    # Force use selector event loop on windows
+    return asyncio.SelectorEventLoop()
```
adettori, line 118: Have you tried something like 
`if sys.platform.startswith('win32'):
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())` 
and use asyncio.run() in create() instead? Seems like it would abstract the platform better that way

g4f/Provider/base_provider.py
```diff
@@ -70,10 +72,7 @@ def create_completion(
         stream: bool = True,
         **kwargs
     ) -> CreateResult:
-        check_running_loop()
-
-        # Force use selector event loop on windows
-        loop = asyncio.SelectorEventLoop()
+        loop = get_new_event_loop()
```
adettori, line 75: Did you mean create_event_loop?

adettori (commented):
We are trying to do very similar things in our commits. Both are fine imo, I left some comments in your code, let me know what you think

g4f/__init__.py
```diff
@@ -43,12 +57,20 @@ def create(
...
+    async def create_async(
+        model    : Union[models.Model, str],
+        messages : list[dict[str, str]],
+        provider : Union[type[BaseProvider], None] = None,
+        **kwargs
+    ) -> str:
+        
+        model, provider = get_model_and_provider(model, provider, False)
+
+        if not issubclass(provider, AsyncProvider):
```
adettori, line 73: You can create providers that inherit directly from BaseProvider too here, just need to handle the case.
Ofc you are more familiar with the codebase, so if you are planning to dismiss the use of purely synchronous providers, the change makes sense.

g4f/Provider/base_provider.py
hlohaus: yes

g4f/__init__.py
hlohaus: When we use BaseProvider here, it will still executed synchronous like in create.
I changed already all working providers to "AsyncProvider".

adettori: Thank you for clarifying, all good then!

g4f/Provider/base_provider.py
adettori: Good point, didn't think of that

g4f/Provider/base_provider.py
```diff
@@ -108,14 +118,16 @@ def create_async_generator(
...
-def create_event_loop():
+def create_event_loop() -> SelectorEventLoop:
     # Don't create a new loop in a running loop
-    if asyncio.events._get_running_loop() is not None:
-        raise RuntimeError(
-            'Use "create_async" instead of "create" function in a async loop.')
-
-    # Force use selector event loop on windows
-    return asyncio.SelectorEventLoop()
+    try:
```
adettori, line 123: Probably my bad, I wanted to avoid using undocumented python functions.
If you want to make this work as intended, you probably have to do something like:
`    try:
        asyncio.get_running_loop()
    except RuntimeError:
        return SelectorEventLoop()

    raise RuntimeError(
            'Use "create_async" instead of "create" function in a async loop.')
`

adettori: get_running_loop() throws an exception if there is no event loop.
Therefore you raise your exception only if the try: statement goes smoothly

g4f/Provider/base_provider.py
```diff
@@ -100,6 +108,11 @@ def create_async_generator(
     ) -> AsyncGenerator:
         raise NotImplementedError()
 
+# Don't create a new loop in a running loop
+def check_running_loop():
+    if asyncio.events._get_running_loop() is not None:
```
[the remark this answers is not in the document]
hlohaus: changed

g4f/Provider/base_provider.py
hlohaus: I changed it

adettori: Looks good to me then

xtekky:
Looks very good !, nice job.

johnd0e:
```
  File "D:\AI\gpt4free\g4f\Provider\ChatgptLogin.py", line 59, in create_async
    raise RuntimeError(data["msg"])
RuntimeError: Your account is not active, please check your billing details on our website.
```

hlohaus:
ChatgptLogin is currently not working. The website too

Issue #915, opened 2023-09-18
RuntimeError: Cannot run the event loop while another loop is running


damarega:
**Bug description**
#Code:
import g4f

print(g4f.Provider.Ails.params)  # supported args

response = g4f.ChatCompletion.create(
    model="gpt-3.5-turbo",
    messages=[{"role": "user", "content": "Hello world"}],
    stream=True,
)

for message in response:
    print(message)

#Error:
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
[<ipython-input-27-e33ca77624cc>](https://localhost:8080/#) in <cell line: 1>()
----> 1 for message in response:
      2     print(message)

3 frames
[/usr/local/lib/python3.10/dist-packages/g4f/Provider/base_provider.py](https://localhost:8080/#) in create_completion(cls, model, messages, stream, **kwargs)
     91         **kwargs
     92     ) -> CreateResult:
---> 93         yield from run_generator(cls.create_async_generator(model, messages, stream=stream, **kwargs))
     94 
     95     @classmethod

[/usr/local/lib/python3.10/dist-packages/g4f/Provider/base_provider.py](https://localhost:8080/#) in run_generator(generator)
    120     while True:
    121         try:
--> 122             yield loop.run_until_complete(gen.__anext__())
    123 
    124         except StopAsyncIteration:

[/usr/lib/python3.10/asyncio/base_events.py](https://localhost:8080/#) in run_until_complete(self, future)
    623         """
    624         self._check_closed()
--> 625         self._check_running()
    626 
    627         new_task = not futures.isfuture(future)

[/usr/lib/python3.10/asyncio/base_events.py](https://localhost:8080/#) in _check_running(self)
    584             raise RuntimeError('This event loop is already running')
    585         if events._get_running_loop() is not None:
--> 586             raise RuntimeError(
    587                 'Cannot run the event loop while another loop is running')
    588 

RuntimeError: Cannot run the event loop while another loop is running

**Environment**
- python version 3.10 (Google Colab 18 September 2023 version)

bitsnaps:
I got the same issue on colab.

PR #926, opened 2023-09-20
Introduce an async friendly ChatCompletion method

adettori:
Hi, I tried to solve the issues mentioned in #915 since I was facing it myself and using this library together with other async ones got tricky.
I tested both the new async_create and create methods, they seem to work as intended.
I'm open to any feeback, lmk

### DIFF
--- a/g4f/Provider/base_provider.py
+++ b/g4f/Provider/base_provider.py
@@ -39,14 +39,15 @@ class BaseProvider(ABC):
 
 class AsyncProvider(BaseProvider):
     @classmethod
-    def create_completion(
+    async def create_completion(
         cls,
         model: str,
         messages: list[dict[str, str]],
         stream: bool = False,
         **kwargs
     ) -> CreateResult:
-        yield asyncio.run(cls.create_async(model, messages, **kwargs))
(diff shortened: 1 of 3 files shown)


g4f/Provider/base_provider.py
```diff
@@ -60,26 +61,22 @@ class AsyncGeneratorProvider(AsyncProvider):
...
-                try:
-                    yield loop.run_until_complete(gen.__anext__())
-                except StopAsyncIteration:
-                    break
-        finally:
-            asyncio.set_event_loop(None)
-            loop.close()
+
+        while True:
+            generator  = cls.create_async_generator(model, messages, stream=stream, **kwargs)
```
hlohaus, line 73: You should call create_async_generator once

adettori: Thank you for pointing that out, I missed it, should be fixed now.

hlohaus:
Nice. I try to fix the event loop error on windows.
Also i added a "create_async" function in ChatCompletion for you in my pr.
https://github.com/xtekky/gpt4free/pull/924
excited for your feedback

PR #929, opened 2023-09-21
Update DeepAi.py

VadimBoev:
DeepAi changed the link an hour ago


VadimBoev:
this is no longer particularly relevant, they require authorization (google account or something similar)

PR #933, opened 2023-09-23
Improve Vercel Provider:

hlohaus:
- Fix endless loop
 - Add proxy, async support
 - Add default model
 Fix HuggingChat Provider


xtekky:
People shouldn't have to use proxies for Vercel to work. Other than that, everything is perfect.

hlohaus:
When there is a Internal Server Error: A new AntiBotToken helps.
The proxy config is only optional
@xtekky

Issue #935, opened 2023-09-24
[Docker] ModuleNotFoundError: No module named 'transformers' 


docaohuynh:
I've got error when using docker

```
docker compose up

[+] Running 2/2
 ✔ Network gpt4free_default       Created                                  0.2s 
 ✔ Container gpt4free-gpt4free-1  Creat...                                 0.3s 
Attaching to gpt4free-gpt4free-1
gpt4free-gpt4free-1  | Traceback (most recent call last):
gpt4free-gpt4free-1  |   File "/usr/local/lib/python3.9/runpy.py", line 197, in _run_module_as_main
gpt4free-gpt4free-1  |     return _run_code(code, main_globals, None,
gpt4free-gpt4free-1  |   File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
gpt4free-gpt4free-1  |     exec(code, run_globals)
gpt4free-gpt4free-1  |   File "/app/interference/app.py", line 9, in <module>
gpt4free-gpt4free-1  |     from transformers import AutoTokenizer
gpt4free-gpt4free-1  | ModuleNotFoundError: No module named 'transformers'
gpt4free-gpt4free-1 exited with code 1

```
**Bug description**
Follow docker install and got error when docker compose up

PR #936, opened 2023-09-24
#935 [Docker] ModuleNotFoundError: No module named 'transformers' and 'PyExecJS'

r1di:
fixing missing modules


johnd0e:
Really??
Why we need gigantic tensorflow and torch here?

PR #939, opened 2023-09-25
show availables providers

Lin-jun-xiang:
## **Purpose**
Due to the variability of providers, manual adjustments or `try...except...` blocks may be necessary during development. This feature aims to quickly identify the currently available providers by **testing messages** and threads.

## Example

**code:**

```python
import g4f

res = g4f.AvailableProviders.show(threads_num=16, model='gpt-4')
print(res)
```

**results**:

```
['Aichat', 'AItianhu', 'Aivvm', 'ChatBase', 'DeepAi', 'GptGo', 'Vitalentum', 'Wewordle', 'Yqcloud']
```

## Scene

[Document GPT app](https://docgpt-app.streamlit.app/):

Users do not need to manually test whether providers are functioning correctly and can quickly identify the usable providers.

## **Note:**
It is uncertain whether adding `AvailableProviders` to `g4f/__init__.py` is appropriate.

### DIFF
--- a/g4f/__init__.py
+++ b/g4f/__init__.py
@@ -96,3 +96,39 @@ class Completion:
                                             [{"role": "user", "content": prompt}], stream, **kwargs)
 
         return result if stream else ''.join(result)
+
+
+class AvailableProviders:
+    providers = [getattr(Provider, provider) for provider in Provider.__all__]
+    model = 'gpt-3.5-turbo'
+
+    @classmethod
+    def _test_provider(cls, provider: Provider) -> str:
+        try:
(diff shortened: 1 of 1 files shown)


hlohaus:
We have added the RetryProvider. You can simply do:

```python
import g4f

    model=g4f.models.gpt_3_turbo,
    messages=[{"role": "user", "content": "hi"}],
)

# or some more providers:

    model=g4f.models.default,
    messages=[{"role": "user", "content": "hi"}],
)

print(response)
```

This is better then call all Models, even they need auth or does not support GPT.

Issue #942, opened 2023-09-26
Probably unnecessary dependencies


RockChinQ:
**Bug description**
Hi there, I just made a project which depends gpt4free, but when I build a docker image for it, I found it's extremly large (up to 10GB), and I found torch, tensorflow and transformers depended by g4f.  
In my view, such a reverse engineering lib will only need requests, websockets or other network-related libs, is these DL related libs unnecessary and can I uninstall them?

**Environement**
- base image: python:3.10.13-bullseye
- python 3.10.13
- location: China, host loc: Eastern Asia

**Additional context**
- dockerfile: https://github.com/RockChinQ/free-one-api/blob/main/Dockerfile
- docker image: rockchin/free-one-api

iaston:
agree +1

RockChinQ:
After uninstalling these dependencies and delete dir starts with nvidia* in the site-package dir, g4f still work well. So I guess these packages were added accidently?

PR #944, opened 2023-09-26
Chore: move ML related deps to interference dir

RockChinQ:
## Intention

- Move `transformers` `tensorflow` `torch` from `/requirements.txt` to `/interference/requirements.txt`

## Reason

After search through each `.py` file, I found `transformers` only imported by `interference/app.py`, this file is http server, and its deps should be save in `interference/requirements.txt`. Also, saving these deps in `/requirements.txt` may cause people who only use reverse engineering features installing unnecessary packages(which are extremly large).

## Related Issue

#942

### DIFF
--- a/interference/requirements.txt
+++ b/interference/requirements.txt
@@ -1,2 +1,5 @@
 flask_cors
 watchdog~=3.0.0
+transformers
+tensorflow
+torch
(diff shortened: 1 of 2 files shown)


RockChinQ:
Also, `flask` and `flask-cors` may need to be moved to `/interference/requirements.txt` too.

Issue #989, opened 2023-10-05
NameError: name 'AItianhu' is not defined


v67bruno:
**Bug description**
`g4f/models.py", line 65, in <module> DeepAi, ChatgptLogin, ChatgptAi, Aivvm, GptGo, AItianhu, Aichat, AItianhuSpace, Myshell, Aibn, ChatForAi, FreeGpt, Ylokh, Cromicle`
`NameError: name 'AItianhu' is not defined`

**Environement**
- python 3.10.9
- g4f 1.4.9

KirovClassBattlecruiser:
same issue, im on python 3.11.5 on windows 10, having installed g4f 1.4.9 with python -m pip install -U g4f!

KirovClassBattlecruiser:
https://github.com/xtekky/gpt4free/pull/991 fixes it

Josh-XT:
@xtekky can you do a release so that this fix is applied to the Python package on Pypi?  The current version is broken.

xtekky:
fixed

PR #991, opened 2023-10-05
fix import in models

ostix360:
imort AItianhu and AItianhuSpace in models.py needed line 65


KirovClassBattlecruiser:
fixes https://github.com/xtekky/gpt4free/issues/989
its usable again!

PR #1008, opened 2023-10-08
Give Yqcloud system prompt & user id arguments

kggn:


hlohaus:
Thank you. In other providers we have "system_message" not "system_prompt".
And we use lower_underscore in Python: userid -> user_id

PR #1009, opened 2023-10-08
add cookies by automatically fetching from chrome browser and skip th…

hdsz25:
…e origin user submitted information in the returned message for OpenaiChat


hlohaus:
Hello, thank you.

What are you doing?
Does  `get_cookies` from default not working?
Why you remove the custom timeout config?
Why you set a cookie header and not use the "cookies" argument?
Why you are not use and change the `get_browser` helper?

PR #1051, opened 2023-10-12
Unify g4f tools (gui and api) into one CLI

ahobsonsayers:
This PR creates a unified g4f CLI that is able to run the gui or the interference api. This CLI has also been added to the setup.py console_scripts so that the gui and api are easily runnable once gpt4free is installed via pip.

Notes:
- I have added a section in the readme mentioning that g4f can also be installed via pipx if a user wants to only use the CLI tools and not the pacakge as a library.
- Moved files from the `interference` folder to the `api` folder to more accuratly represent the name of the CLI tool
- I have removed the duplicated interference code from the etc folder - it is uneeded

Please let me know if there are any modifications you would like to make


g4f/api/__init__.py
ahobsonsayers, line 1: This entire file is copied from the `g4f/interference` folder with the embeddings being added from the `etc/interference/app.py` file

xtekky:
looks good to me

Issue #1014, opened 2023-10-08
How i can dont use Bing?


tima24312:
Hello, I am completely bad at programming and very bad at coding. How can I disable the use of provider "Bing" when replying via GPT-4 in the web interface? I have installed this via "pip install -U g4f". Or how can I set up a specific provider

xtekky:
setting a specific provider is not available at the moment on the web interface

tima24312:
> setting a specific provider is not available at the moment on the web interface

Could there be an opportunity to do so in the near future?

PR #1064, opened 2023-10-13
feat: ignore providers

Lin-jun-xiang:
According to #1014 , add a `ignored` parameter which can let users choose **ignore specific providers**.

## Example

```python

# normal
    model='gpt-3.5-turbo',
    messages=[{"role": "user", "content": "hello"}],
    ignored=["Ylokh", "GptGo", "AItianhu", "Aibn", "Myshell", "FreeGpt"] # Ignore these providers
)

print(response)

# async
async def test():
    response = await g4f.ChatCompletion.create_async(
        model='gpt-3.5-turbo',
        messages=[{"role": "user", "content": "hello"}],
        ignored=["Ylokh", "GptGo", "AItianhu", "Aibn", "Myshell", "FreeGpt"] # Ignore these providers
    )
    return response

async def run_all():
    calls = [
        test()
    ]
    res = await asyncio.gather(*calls)
    return res

print(asyncio.run(run_all()))
```

If the feature is needed, perhaps someone could extend it and implement it in the **web UI**?
(As I'm not proficient in web development.)

### DIFF
--- a/g4f/__init__.py
+++ b/g4f/__init__.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 from requests import get
 from g4f.models import Model, ModelUtils
-from .Provider import BaseProvider
-from .typing import Messages, CreateResult, Union
+from .Provider import BaseProvider, RetryProvider
+from .typing import Messages, CreateResult, Union, List
 from .debug import logging
 
 version = '0.1.6.2'
 version_check = True
+
 
 def check_pypi_version() -> None:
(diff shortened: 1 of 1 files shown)


xtekky:
the users can also just use a list of providers
```py
from g4f.Provider import a, b, c
from random import choice

ChatCompletion.create(..., provider=choice([a, b, c])
```

but this is a feature after all, I'll be merging

PR #1066, opened 2023-10-13
Add stage to CI/CD for publishing image to Github Container Registry

Niapoll:
I found of that the **CI/CD workflow** for publishing image to **DockerHub** was removed. Unfortunately, I didn't find why :( (legal purposes?):
```bash
* 08291e8 - Delete ci.yml (10 weeks ago) <xtekky>
* 9b69fc6 - Add docker image ci support (6 months ago) <wengchaoxi>
```
So, I think we can use **Github Container Registry** for these purposes, aren’t we? I have created workflow which publish a new image to **Github registry** every time when a tag has been pushed.

UPD: Also I found of that **Docker compose** launch only **API** service. I have added the **GUI** service to **Docker compose**.

### DIFF
--- a/.github/workflows/publish-workflow.yaml
+++ b/.github/workflows/publish-workflow.yaml
@@ -1,0 +1,33 @@
+name: Publish Docker image
+
+on:
+  push:
+    tags:
+      - '**'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Buildx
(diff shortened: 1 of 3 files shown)


xtekky:
I removed it because It was causing installation issues, but I am happy to add this feature again

xtekky:
@Niapoll looks like there is some issues publishing the package, I have set up the right credentials and acess token via repository secrets etc...

image

any clue on how to resolve this ?

PR #1086, opened 2023-10-16
Add support for openai.completion.create in __init__.py

chatgpt-tricks:


chatgpt-tricks:
I removed the retry part. 
btw do you know if the retry in the create function tries again with the same model or with a different one, because sometimes I get a lot of the same errors.

xtekky:
There is a lot to be improve here, stream, use Completion instead of ChatCompletion etc...

PR #1106, opened 2023-10-20
Fix bugs

catmeowjiao:
Here's what I've done:

1. Deleted MANIFEST.in from .gitignore and added MANIFEST.in to the project to fix the issue where the build directory created by python setup.py build doesn't include g4f/gui/client and g4f/gui/server.

2. Changed /chat/completions in the API to /v1/chat/completions to ensure compatibility with the OpenAI API.

As for what I couldn't achieve:

1. Adding proxy parameters to the API and GUI. (Challenge: How can the program handling the parameters inform the program using g4f.ChatCompletion.create about the proxy address?)

By meowjiao

Using ChatGPT from Chinese translate to English

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,6 @@ lab.js
 bing.py
 bing2.py
 .DS_Store
-MANIFEST.in
 lab/*
 lab
 tstt.py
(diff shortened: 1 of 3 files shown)


xtekky:
adding /v1 to the url is not necessary, as it will then have to be added to the prefix anyways:

```py
openai.api_key = ""
openai.api_base = "http://localhost:1337/v1"
```

xtekky:
I completely updated the api, you might check the new version and update your pull request

PR #1118, opened 2023-10-21
Add FakeGpt Provider

hlohaus:
Add FakeGpt, Hashnode Provider
Update used providers in models
Increase timeout in RetryProvider with kwargs
Enable Liaobots and ChatForAi again
Fix Yqcloud and MyShell Provider


Luneye:
Thanks for being so committed to developing this project man, you're such a dedicated soldier lol

PR #1120, opened 2023-10-22
Added code to ignore certain providers in the API

AndPim4912:


AndPim4912:
#1120

hlohaus:
Why you don't add it to Api.__init__?
ApiInterface isn't a interface.
It is a decorator.

PR #1157, opened 2023-10-25
Trying to add context for gui

Biflez48:
**(improved)**
I really did not like that gui does not remember previous messages after restarting the page. Therefore, I made an attempt to add a context, especially since there is everything necessary for this.
As far as I am not very good at python and javascript, I have a problem. I managed to collect the message history and send it to the chat.v1.js file. Also sent this message history through stream () in the backend.py file, just as it was done in README. Now, when you try to send a request, or get the result, an error of the form 
image
 crawls out. In this case, an 
image
 message appears on the gui page itself. There are ideas and suggestions that may not be the way and how to fix it?

### DIFF
--- a/g4f/gui/client/js/chat.v1.js
+++ b/g4f/gui/client/js/chat.v1.js
@@ -31,12 +31,48 @@ const handle_ask = async () => {
     message_input.style.height = `80px`;
     message_input.focus();
 
+    // start parse
+
+    let txtMsgs = [];
+    let UorA = true;
+    const divTags = document.getElementsByClassName("content");
+    for(let i=0;i<divTags.length;i++){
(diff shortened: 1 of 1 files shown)


Biflez48:
Urgent inclusion!!! Removed the excess from the compilation of the message and now there is a working context and memorization of past messages!!! 
image

catmeowjiao:
@Biflez48 You need to process when the role is system

Biflez48:
> @Biflez48 You need to process when the role is system

@catmeowjiao And if you mean errors, then I agree with you. But this code is an outline. At the moment I am completely reworking the entire code to improve this context

catmeowjiao:
Sorry, I thought the GUI could customize system messages.
~~I am an idiot.~~

PR #1213, opened 2023-11-04
Make the API use FastAPI instead of Flask

thatlukinhasguy1:
- Made the API ``__init__`` module use FastAPI.
- Made the CLI mode compatible with the new API code.
- Made the ``run`` module compatible with the new API code.


VienDC:
This would be interesting and effective

xtekky:
hey sorry for the delayed merge, but thanks !
Lets hope that fasapi will bring some performance improvements to the api.

PR #1235, opened 2023-11-11
Authentication

thatlukinhasguy1:
Added an optional parameter in the CLI G4F app for API key-based authentication.


hlohaus:
You should use compare diggest for secure comparison.
Here's an example:

```python
import secrets

str1 = "example1"
str2 = "example2"

result = secrets.compare_digest(str1, str2)

if result:
    print("Strings are equal.")
else:
    print("Strings are not equal.")
```
Also please solve conflicts.

PR #1250, opened 2023-11-14
stream check workaround & fix

kacperkwapisz:
fixed the line where it checks if stream is requested. I made the change because before it ignored what the user typed because there was a invalid conversion from the boolean to string where the if stream clause would always be true. Simple workaround :)

### DIFF
--- a/g4f/api/__init__.py
+++ b/g4f/api/__init__.py
@@ -78,7 +78,7 @@ class Api:
                 item_data['messages'] = ast.literal_eval(item_data.get('messages'))
 
             model = item_data.get('model')
-            stream = item_data.get('stream')
+            stream = True if item_data.get("stream") == "True" else False
             messages = item_data.get('messages')
 
             try:


hlohaus:
You can pass a boolean in the json struct. Your Changes works only with the "False" string.

kacperkwapisz:
> You can pass a boolean in the json struct. Your Changes works only with the "True" string.

did you even look into the code? i know what i did and what i did is correct because the struct ist boolean but then idk why it converts it into string. show a better solution rather than telling me something that i did with purpose

hlohaus:
> > You can pass a boolean in the json struct. Your Changes works only with the "True" string.
> 
> did you even look into the code? i know what i did and what i did is correct because the struct ist boolean but then idk why it converts it into string. show a better solution rather than telling me something that i did with purpose

Yes it is okay. I'll merge it at your risk.

PR #1258, opened 2023-11-16
Add back crutch of incorrect responses to ChatBase

AndPim4912:
70% of responses content order to contact support instead of answer. It must not be treated as normal answer and raise ValueError.


hlohaus:
Thank you

PR #1272, opened 2023-11-20
Fix tupo in GptGo

AndPim4912:
: prevent loading


AndPim4912:
We need unittest autocheck.

PR #1281, opened 2023-11-22
Fix Phind and GptGo Provider

hlohaus:


SilverMarcs:
After this merge, Phind is no longer working for me. Previously, if I were hosting gpt4free inference server on some computer, it used to open a selenium chrome window, submit request and send back response to the API caller client. Now it doesnt do anything at all (does does show a 200 request on the console).
What was wrong with the phind provider to begin with?

xtekky:
> After this merge, Phind is no longer working for me. Previously, if I were hosting gpt4free inference server on some computer, it used to open a selenium chrome window, submit request and send back response to the API caller client. Now it doesnt do anything at all (does does show a 200 request on the console).
> 
> What was wrong with the phind provider to begin with?

It was using the selenium driver, now it's pure http requests - so faster, I can take a look and see if it really broke it.

SilverMarcs:
I think even in the previous version, it was always defaulting to phind model even if gpt-4 was passed. On the chrome window, it showed gpt-4 model was selected in the model picker button but the output was generated using phind model (the site shows which model was actually used for the output). dunno if this can be fixed or if the site is somehow preventing gpt-4 through selenium. 

As with the new http request version, it does not indeed work for me

PR #1291, opened 2023-11-26
Added new provider PI (Hacky way to use)

madonchik123:
Context Limit is 1k for input
And messages is only used as prompt
Chat history is stored in conversation which you need to save if you want to save chat history


xtekky:
Hi, will check this soon !

hlohaus:
Thank you. You use AsyncGeneratorProvider but the requests aren't async

PR #1352, opened 2023-12-15
HuggingChat: Strip leading whitespace from the first token in the stream

nullstreak:
For some reason first token from the stream on HuggingChat always starts with a whitespace. This commit strips the leading whitespace from the first token in the stream to fix this issue.


hlohaus:
Thank you

PR #1360, opened 2023-12-17
Bing: add parameter to enable/disable web search

nullstreak:
Web search is disabled by default and can be enabled by passing `web_search = True` into `g4f.ChatCompletion.create`


hlohaus:
thank you

PR #1364, opened 2023-12-18
Fix Bing provider Cookie issue 2

eminemkun:
#1356 

I got the same error again.

```
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: ', url=URL('https://www.bing.com/turing/conversation/create?bundleVersion=1.1199.4') 
```

SRCHHPGUSR needs to include the current UNIXTIME. Probably related to the time limit.

### DIFF
--- a/g4f/Provider/Bing.py
+++ b/g4f/Provider/Bing.py
@@ -10,6 +10,7 @@ import base64
 import numpy as np
 import uuid
 import urllib.parse
+import time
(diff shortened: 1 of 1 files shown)


iG8R:
Thanks a lot! Now it works.

hlohaus:
thank you

PR #1398, opened 2023-12-26
Fix Phind provider

SilverMarcs:
- Adds challenge parameter in request body to match requests made on the Phind web app
- Prevents new <PHIND_INDICATOR> tag from being added to response body

Not really sure what the challenge parameter does but the value I added is roughly what they use for all the requests on phind web ui


hlohaus:
thank you

PR #1430, opened 2024-01-04
Fix module 'g4f.debug' has no attribute 'get_version'

ramonvc:
Fix error `module 'g4f.debug' has no attribute 'get_version' ` 

that occurs when running the API directly from the local repository using `python -m g4f.api.run`.

image


hlohaus:
Thank you

PR #1431, opened 2024-01-04
Update openai example

ramonvc:
After the release of version 1.0.0 of the openai python package, openai.ChatCompletion is no longer supported.

This change updates the openai example to the latest version of the package.


hlohaus:
Thank you

PR #1461, opened 2024-01-13
A slight cleanup to the project

Akash98Sky:
* Cleanup unused dependencies
* remove packages, that are not a direct dependency to this project, from requirements.txt
* no need to just numpy for simple sqrt operation on numbers
* write code on Python than using a JS compiler


Akash98Sky:
I feel there are a few more dependencies that we can get rid of, which will make this project lightweight and faster.

1. **PyExecJS** - This is used only by Vercel provided. It is used to compile some lines of JS code. Can we not get rid of this completely, and write some equivalent python functionality?
2. **requests**/**curl_cffi**/**aiohttp** - All these are used as http client library in this project. Can we not use a single http client library that serves all our purposes?

hlohaus:
Thank you

PR #1480, opened 2024-01-18
Llama2.py result decode error：UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 6

diaodeng:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 6


hlohaus:
thank you

PR #1523, opened 2024-01-28
Fix chromedriver path

u66u:
I had some problems with the hardcoded webdriver path, so:

1. Discover chromedriver dynamically with "which". 
2. Use "finally" in WebDriverSession.__exit__


hlohaus:
Thank you

PR #1529, opened 2024-01-30
Fix DeepInfra: Model is not supported

nullstreak:
Last commit broke DeepInfra


nullstreak:
Btw maybe it's better to hardcode supported models instead of spamming API on every inference?

hlohaus:
Maybe, but it is cached. Thank you

PR #1532, opened 2024-01-31
Added aiohttp_socks package into Dockerfile

delfer:
Required to run Docker container with option -e G4F_PROXY=...


hlohaus:
I added it already, but doesn't released it. I created now a Release.

PR #1601, opened 2024-02-18
Move GeminiProChat to a new provider #1596

### LINKED ISSUE #1596: Gemini prochat provider
It seems that geminiprochat.com is down, so i found an other provider : https://gemini-chatbot-sigma.vercel.app/

I try to reverse it but it seems that the signature type has been modified in comparison to geminiprochat.com...

piotr25691:
Since GeminiProChat is down I decided to take the existing provider file and adapt it for https://gemini-chatbot-sigma.vercel.app/ which is a provider with the same backend and frontend, just a different name. That one actually works.

Merging this would close #1596

### DIFF
--- a/g4f/Provider/GeminiChatbotSigma.py
+++ b/g4f/Provider/GeminiChatbotSigma.py
@@ -8,8 +8,8 @@ from ..typing import AsyncResult, Messages
 from .base_provider import AsyncGeneratorProvider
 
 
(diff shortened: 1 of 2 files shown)


piotr25691:
didnt work

hlohaus:
Thank you

hlohaus:
Gemini
GeminiPro
and
GeminiProChat
are working ^^

PR #1634, opened 2024-02-26
Change uses of whitelist to allowlist in translate_readme.py.

xwxtwd:
1.Change uses of whitelist to allowlist in translate_readme.py.
2.Change uses of blacklist to blocklist in translate_readme.py.
3.Change uses of blacklisted to blocklisted in translate_readme.py.


hlohaus:
Thank you
Are you using translate_readme.py for your project?

PR #1663, opened 2024-03-08
Fix TypeError: 'Api' object is not callable

ramonvc:
Run interference API from PyPi package results in a _TypeError: 'Api' object is not callable_ when attempting to make a request.

errorAPI

The adjustment ensures that the bug is fixed while keeping the same functionalities, including the _use_colors_ or run interference API from repo

done


hlohaus:
Thank you. Very fine

PR #1667, opened 2024-03-08
Expire cache, Fix multiple websocket conversations in OpenaiChat

hlohaus:
Map system messages to user messages in GeminiPro


10121984 (approved):
g4f/gui/client/js/chat.v1.js

PR #1670, opened 2024-03-09
~ gui system prompt spacing

xtekky:
The spacing was misaligned, a little increase in the width fixes it

before:
image

after:
image

on chrome, mac air m2


hlohaus:
@xtekky why 59 and not 60? Also I wanted to add a min height.

xtekky:
> @xtekky why 59 and not 60? Also I wanted to add a min height.

well it looked better with 59, 60 may work aswell

hlohaus:
Ok

PR #1704, opened 2024-03-14
Error checking for Vercel

Abobo8964:
Sorry I submitted three times (because I tried to submit directly through the file, but it seemed to fail), this PR fixes Vercel's UnicodeDecodeError error so that the answer no longer interrupts


hlohaus:
Thank you 😌

PR #1712, opened 2024-03-16
feat: docker add arm64 support

exfly:
When I started this project with Docker on my MacBook Pro M1 chip, I found that Chrome would get an error, probably because Chrome only supports the 4K Pagesize AMD64 platform, not Arm64. After investigating, the current `seleniumnode/chrome` image used by the dockerfile only supports amd64. The selenium community is already pushing for the arm64 merge https://github.com/SeleniumHQ/docker-selenium/issues/1847. For the time being, we are using the base image `seleniarm/node-chromium` that supports both arm64 and amd64 to ensure dual-architecture support. 

Currently, arm64 images support Pagesize 4k kernels, and there is no support plan for Pagesize 64k in the short term (Chrome does not support 64k Pagezie).

### DIFF
--- a/.github/workflows/publish-workflow.yaml
+++ b/.github/workflows/publish-workflow.yaml
@@ -11,6 +11,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
 
       - name: Get metadata for Docker
         id: metadata
@@ -38,6 +42,7 @@ jobs:
         with:
           context: .
           file: docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.metadata.outputs.tags }}
           labels: ${{ steps.metadata.outputs.labels }}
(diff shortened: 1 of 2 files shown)


hlohaus:
Hello, thank you.
You replaced only amd64 with arm64.
You should try to add support for both.
Maybe you can create a new Docker file.

exfly:
image

Already supports arm64 and amd64

.github/workflows/publish-workflow.yaml
```diff
@@ -38,6 +42,7 @@ jobs:
         with:
           context: .
           file: docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
```
exfly, line 45: Both support amd64 and arm64

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM selenium/node-chrome
+FROM seleniarm/node-chromium
```
exfly, line 1: Both support amd64 and arm64

exfly:
> Hello, thank you.
> You replaced only amd64 with arm64.
> You should try to add support for both.
> Maybe you can create a new Docker file.

the base image seleniarm/node-chromium are supports both arm64 and amd64

hlohaus:
Okay, I understand. Thank you

exfly:
It seems that an image has been built to support the two architectures arm64 and arm64.

Issue #1746, opened 2024-03-23
An error occured: You: ResponseStatusError: Response 401: DFP telemetry ID not set   (Opus and Sonnet model)


Woukim:
![image](https://github.com/xtekky/gpt4free/assets/138323639/2992375c-518c-42bc-851c-e6657f2ccd5e)

**Bug description**
 just start chat and got an error

**Environment**
last docker

hlohaus:
You need a account for all models except gpt-3.5. The account creation may have changed. You.com is the only provider, who have a auto account creation in g4f. I don't like to create and new prov of the account creation. You can use a free account only 5x times. Maybe we can separate the account creation and the requests to the models?

kirill670:
![image](https://github.com/xtekky/gpt4free/assets/51964569/5cda8633-75da-4682-927e-3eff70699508)
Same thing with You.

rqi14:
Same problem

BokDan93:
I agree, I have this problem too. Already tried changing the ip and server

rqi14:
Any possibility to fix this?

AksLolCoding:
i got a fix to work by creating an account in the browser and then loading those into g4f
i'm making a pr to implement this
but it requires you to be signed in already

AksLolCoding:
I fixed the issue and created a [pull request](https://github.com/xtekky/gpt4free/pull/1794)
I tested that it works

rqi14:
> I fixed the issue and created a [pull request](https://github.com/xtekky/gpt4free/pull/1794) I tested that it works

Thank you very much!

Issue #1739, opened 2024-03-22
create_cookies error in You provider


sjrsjz:
log:
Using You provider and claude-3-sonnet model
Traceback (most recent call last):
  File "E:\mirai\AliceBot\AliceBot.py", line 56, in <module>
    test=g4f.ChatCompletion().create(
  File "f:\Python_3.10.9\lib\site-packages\g4f\__init__.py", line 149, in create
    return result if stream else ''.join([str(chunk) for chunk in result])
  File "f:\Python_3.10.9\lib\site-packages\g4f\__init__.py", line 149, in <listcomp>
    return result if stream else ''.join([str(chunk) for chunk in result])
  File "f:\Python_3.10.9\lib\site-packages\g4f\providers\base_provider.py", line 206, in create_completion
    yield loop.run_until_complete(await_callback(gen.__anext__))
  File "f:\Python_3.10.9\lib\site-packages\nest_asyncio.py", line 98, in run_until_complete
    return f.result()
  File "f:\Python_3.10.9\lib\asyncio\futures.py", line 201, in result
    raise self._exception.with_traceback(self._exception_tb)
  File "f:\Python_3.10.9\lib\asyncio\tasks.py", line 234, in __step
    result = coro.throw(exc)
  File "f:\Python_3.10.9\lib\site-packages\g4f\providers\base_provider.py", line 202, in await_callback
    return await callback()
  File "f:\Python_3.10.9\lib\site-packages\g4f\Provider\You.py", line 69, in create_async_generator
    cookies = await cls.get_cookies(session) if chat_mode != "default" else None
  File "f:\Python_3.10.9\lib\site-packages\g4f\Provider\You.py", line 145, in get_cookies
    cls._cookies = await cls.create_cookies(client)
  File "f:\Python_3.10.9\lib\site-packages\g4f\Provider\You.py", line 185, in create_cookies
    await raise_for_status(response)
  File "f:\Python_3.10.9\lib\site-packages\g4f\requests\raise_for_status.py", line 23, in raise_for_status_async
    raise ResponseStatusError(f"Response {response.status}: {message}")
g4f.errors.ResponseStatusError: Response 401: DFP telemetry ID not set

However when use gpt-3.5 it won't throw error

TheColonists:
I am also having the 401 error.
```
Using RetryProvider provider and gpt-4 model
Using Raycast provider
Using ChatgptAi provider
ChatgptAi: RateLimitError: Rate limit reached
Using You provider
[...]

PR #1794, opened 2024-04-06
Update You.py

AksLolCoding:
fixed [this issue](https://github.com/xtekky/gpt4free/issues/1746) and [this issue](https://github.com/xtekky/gpt4free/issues/1739)


hlohaus:
I discovered this alternative solution today too. Initially, I began constructing a generator, however, I subsequently employed a random UUID, which proved to be equally effective.

PR #1827, opened 2024-04-12
Update DeepInfra models

nullstreak:
**codellama/CodeLlama-34b-Instruct-hf** - Removed, no longer available for free
**mistralai/Mistral-7B-Instruct-v0.1** - Removed, no longer available for free
**jondurbin/airoboros-l2-70b-gpt4-1.4.1** - Removed, not available anymore

**mistralai/Mistral-7B-Instruct-v0.2** - Added new model

**mistralai/Mixtral-8x22B-v0.1** - Added, but commented out because it doesn't work (yet?) with OpenAI async_generator. Perhaps we need to change the implementation in DeepInfra.py for it to work.

### DIFF
--- a/g4f/models.py
+++ b/g4f/models.py
@@ -127,7 +127,7 @@ llama2_70b = Model(
 codellama_34b_instruct = Model(
     name          = "codellama/CodeLlama-34b-Instruct-hf",
     base_provider = "meta",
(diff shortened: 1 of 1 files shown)


hlohaus:
Thank you!

PR #1860, opened 2024-04-19
Updated model aliases (added -instruct to models)

pratham-darooka:
I overlooked a bug in models.py while testing. Fixed with new model name.


hlohaus:
You are to late, i merged already o other pr with this fixes.
Thank you

PR #1914, opened 2024-05-02
add Ollama as a Provider

hdsz25:
web gui can use Ollama as a provider and also added below models for user to select.
phi3,qwen,mistral,gemma,codellama

In order to use Ollama, one needs to install ollama and download LLMs to make it work!
https://ollama.com/


hlohaus:
Hey, can you add a get_models function and extend the provider from the OpenAI provider? That way, there's less copy and paste, and the user can choose from the supported models.

hdsz25:
> Hey, can you add a get_models function and extend the provider from the OpenAI provider? That way, there's less copy and paste, and the user can choose from the supported models.

Hi, When I am trying to avoid the duplication I coded as below:
`from .Openai import Openai
from ...typing import Union, Optional, AsyncResult, Messages

class Ollama(Openai):
    label = "Ollama"
    url = "http://localhost:11434"
    api_base = "http://localhost:11434/v1/"

    @classmethod
    async def create_async_generator(
        cls,
        model: str,
        messages: Messages,
        api_key: str = 'Ollama',  # You can set a default API key for Ollama if needed
        **kwargs
    ) -> AsyncResult:
        # Call the parent class's create_async_generator method with the updated api_base
        return await super().create_async_generator(
            model=model,
            messages=messages,
            api_key=api_key,  # Use the provided api_key or the default one for Ollama
            api_base=cls.api_base,  # Use the new api_base from the Ollama class
            **kwargs
        )
`
but it returns error as below:
image

hlohaus:
Remove the async before the def definition.

hlohaus:
I added Ollama for you:
https://github.com/xtekky/gpt4free/pull/1925

hdsz25:
Thanks so much!

PR #1926, opened 2024-05-06
Fix Ollama not working in g4f api

hdsz25:
it's not working in api, please merge only below files to fix this bug:
g4f/providers/base_provider.py
g4f/cli.py
g4f/api/__init__.py
g4f/Provider/Ollama.py

Also set default model for Ollama to use phi3:latest if available.

Thanks!


hlohaus:
Hey, I fixed AppConfig in the latest release.
I added args support for the model and provider in the master branch. I could add phi as the default model. Nothing else should be changed. ?

hdsz25:
> Hey, I fixed AppConfig in the latest release. I added args support for the model and provider in the master branch. I could add phi as the default model. Nothing else should be changed. ?

Thanks for the quick reply. The api with parameters give below error:
image

hlohaus:
Hey, I just updated the Bing provider in the latest version.

PR #1958, opened 2024-05-15
Add needs auth to provierds, Add PerplexityApi provider

hlohaus:
Add proxy support to nodriver


vrashad:
> Add proxy support to nodriver

I solved this issue proxying worker.exe, chrome.exe, python.exe using Proxifier
But it would really be great if you could install a proxy through code

PR #1966, opened 2024-05-17
Improve compatibility of openai-node

kokarare1212:
When using the openai-node library,
the following error occurs in the code below due to a missing implementation of OpenAI's API.
```javascript
import OpenAI from 'openai';

const openai = new OpenAI({
    apiKey: "gpt4free",
    baseURL: "http:///gpt4free-api.url/v1",
});

stream = await openai.beta.chat.completions
    .stream({
      "model":"gpt-3.5-turbo",
      "stream":true,
      "messages":[{"role":"user","content":"Hello!"}],
    });

for await (const chunk of stream) {
    console.log(chunk.choices[0]?.delta?.content || '');
}
```
```
OpenAIError: stream ended without producing a ChatCompletionMessage with role=assistant
```

### DIFF
--- a/g4f/api/__init__.py
+++ b/g4f/api/__init__.py
@@ -120,7 +120,10 @@ class Api:
                 'created': 0,
                 'owned_by': model.base_provider
             } for model_id, model in model_list.items()]
-            return JSONResponse(model_list)
+            return JSONResponse({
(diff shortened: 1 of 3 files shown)


hlohaus:
thank you

Issue #2078, opened 2024-06-23
HuggingChat provider is not working - ResponseStatusError: Response 500


MRGRD56:
### Bug description

When I try to use the HuggingChat provider, having added a cookies/har file, I always get the same error: `An error occurred: HuggingChat: ResponseStatusError: Response 500:`

```
Using HuggingChat provider and CohereForAI/c4ai-command-r-plus model
INFO:werkzeug:192.168.80.1 - - [22/Jun/2024 16:31:48] "POST /backend-api/v2/conversation HTTP/1.1" 200 -
ERROR:root:Response 500: 
Traceback (most recent call last):
  File "/app/g4f/gui/server/api.py", line 177, in _create_response_stream
    for chunk in ChatCompletion.create(**kwargs):
  File "/app/g4f/providers/base_provider.py", line 223, in create_completion
    yield loop.run_until_complete(await_callback(gen.__anext__))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/app/g4f/providers/base_provider.py", line 52, in await_callback
    return await callback()
           ^^^^^^^^^^^^^^^^
  File "/app/g4f/Provider/HuggingChat.py", line 99, in create_async_generator
    await raise_for_status(response)
  File "/app/g4f/requests/raise_for_status.py", line 28, in raise_for_status_async
    raise ResponseStatusError(f"Response {response.status}: {message}")
g4f.errors.ResponseStatusError: Response 500:
```

### Steps to reproduce

1. Put your cookies json file / har file for `huggingface.co` in the `har_and_cookies` directory
2. Run gpt4free in Docker using docker compose
3. Open g4f web ui (using OpenAI compatible API (port `1337`) gives the same error, though)
4. Select this provider: `HuggingChat (Auth)`
5. Select any model, for example `CohereForAI/c4ai-command-r-plus`
6. Send any message to the LLM
7. See the error

### Screenshot

![image](https://github.com/xtekky/gpt4free/assets/35491968/7afaf19b-4af2-4703-8bf3-c4c02eb511fc)

### Environment

- gpt4free version 0.3.2.0 (this git repository, commit `e8f6013d`)
- docker compose
- Ubuntu 22.04.4 LTS x86_64

-----

[...]

PR #2080, opened 2024-06-25
fixed issue #2078

Anurag-Varma:
fixed 500 response error from issue, but still lots of errors in code logic for this provider are present,

will fix them next after getting few doubts clarified

### DIFF
--- a/g4f/Provider/HuggingChat.py
+++ b/g4f/Provider/HuggingChat.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 import requests
-from aiohttp import ClientSession, BaseConnector
+from aiohttp import ClientSession, BaseConnector, FormData
 
 from ..typing import AsyncResult, Messages
 from ..requests.raise_for_status import raise_for_status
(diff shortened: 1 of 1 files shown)


g4f/Provider/HuggingChat.py
```diff
@@ -86,25 +86,36 @@ async def create_async_generator(
...
                     await raise_for_status(response)
                     line = json.loads(line)
                     if "type" not in line:
                         raise RuntimeError(f"Response: {line}")
                     elif line["type"] == "stream":
                         token = line["token"]
-                        if first_token:
-                            token = token.lstrip().replace('\u0000', '')
-                            first_token = False
+                        token = token.lstrip().replace('\u0000', ' ')
```
MRGRD56, line 118: Is it really necessary to use a whitespace here instead of an empty string?

Probably something is wrong because I'm getting lots of whitespace characters using the API

image

but, just replacing `' '` with `''` removes all spaces.

try this

```suggestion
                        token = token.replace('\u0000', '')
```

it works fine, at first glance

Anurag-Varma: Have few , more changes. Will send a new pull request tomorrow.

Anurag-Varma: if i replace  '  ' with ''. Then there is no space between the tokens.

So all the tokens are attached together when i tested it .

MRGRD56: Yes, I've mentioned that :)

have you tried removing the `lstrip()` call like in my comment above?

Anurag-Varma: Yaa, sorry.
I fixed that, its working fine so i deleted the comment

MRGRD56: Nice, thanks

g4f/Provider/HuggingChat.py
```diff
@@ -101,24 +101,37 @@ async def create_async_generator(
...
+                        line = json.loads(line)
+                        if "type" not in line:
+                            raise RuntimeError(f"Response: {line}")
+                        elif line["type"] == "stream":
+                            token = line["token"].replace('\u0000', '')
+                            yield token
+                        elif line["type"] == "finalAnswer":
+                            break
+                else:
+                    content = await response.read()
```
Anurag-Varma, line 123: Added new stream=False feature, where it returns the final whole message after getting all the data from streamreader via await response.read()

g4f/Provider/HuggingChat.py
Anurag-Varma: > image
> 

In the image, how were you able to save and use previous conversations,

I added a new comment in your issue where i sent a SS of how its working for http://127.0.0.1:8080/chat/ & in huggingface chat feature in its website. It's unable to remember the previous messages as they are getting overridden.

xtekky:
Hi, I updated it already

g4f/Provider/HuggingChat.py
MRGRD56: @Anurag-Varma IDK it just worked like that for me, I just used your version :D

I used your fix and it put the entire history in the first message.

That's why I didn't get what you meant, at first. Cause I didn't have that problem, for some reason

But since xtekky has fixed the issue, it doesn't really matter anymore, I suppose. But if this problem is still reproducing for you (I haven't checked the new fix yet), maybe you should open a new issue on it. Try using the clean version from the repository just in case

Issue #2087, opened 2024-06-30
Arkoke token not found in .har file


Baldi89989:
**Bug description**
When using open ai chat, i get a error saying "no arkose token found in .har file" while i used a .har file. How do i fix it?

**Screenshots**
![image](https://github.com/xtekky/gpt4free/assets/97530481/21953b07-1235-4f0f-ab09-786e59056bd6)

**Environment**
Python version: 3.12.1
Location: France

phamxtien:
I got the same

Lorodn4x:
I got the same

gamelist1990:
The problem is due to an update to Chatgpt.com. I tested the code separately and was able to solve the problem, do you want it? But the problem returns a response with banned devices in my environment 😭

gamelist1990:
I stand corrected :not very successful

gamelist1990:
I just tried it and it works fine.

gamelist1990:
## Obtaining the HAR file from ChatGPT

Here's a method to obtain the HAR (HTTP Archive) file from ChatGPT:

1. **Access ChatGPT:** Go to [https://chatgpt.com/](https://chatgpt.com/).
2. **Open Network Tab:** Open your browser's developer tools and navigate to the "Network" tab.
3. **Send a Request:** Interact with the AI by sending a message. Ensure you see a response in the Network tab, indicating a successful request.
4. **Open New Chat:** Click on "New Chat" to start a fresh conversation.
5. **Check Request Count:** Observe the request count in the Network tab. If successful, the count should be approximately between 93 and 104.

At this point, you can right-click on any of the requests and select "Save as HAR" to download the HAR file.

**Note:** The exact request count may vary slightly depending on the specific interaction and browser used.

Lorodn4x:
> ## Obtaining the HAR file from ChatGPT
> Here's a method to obtain the HAR (HTTP Archive) file from ChatGPT:
> 
> 1. **Access ChatGPT:** Go to https://chatgpt.com/.
> 2. **Open Network Tab:** Open your browser's developer tools and navigate to the "Network" tab.
> 3. **Send a Request:** Interact with the AI by sending a message. Ensure you see a response in the Network tab, indicating a successful request.
> 4. **Open New Chat:** Click on "New Chat" to start a fresh conversation.
> 5. **Check Request Count:** Observe the request count in the Network tab. If successful, the count should be approximately between 93 and 104.
> 
[...]

PR #2097, opened 2024-07-05
ChatGPT Reverse Engineering Update: Adapting to Arkose Detection Changes

gamelist1990:
## ChatGPT Reverse Engineering Update: Adapting to Arkose Detection Changes

This pull request addresses a recent change in ChatGPT's backend that modified the way Arkose detection is signaled in API responses.

### Problem:

Previously, the code relied on checking the `requirements["arkose"]["required"]` flag to determine if an Arkose token was necessary. However, ChatGPT has adjusted its API, and this information is now nested within a `text` key within the `requirements` object. This change breaks the existing Arkose detection logic.

### Solution:

This PR modifies the `create_async_generator` function to correctly interpret the updated Arkose detection mechanism.

**Here's how the code is updated:**

**Original code:**

```python
                   as response:
                    cls._update_request_args(session)
                    requirements = await response.json()
                    need_arkose = requirements.get("arkose", {}).get("required")
                    chat_token = requirements["token"]      
```

**Updated code:**

```python
                    as response:
                    cls._update_request_args(session)
                    requirements = await response.json()
                    #Get from text
                    text_data = json.loads(requirements.get("text", "{}")) 
                    need_arkose = text_data.get("turnstile", {}).get("required", False)
                    if need_arkose:
                        arkose_token = text_data.get("turnstile", {}).get("dx")
                    else:
                        need_arkose = requirements.get("arkose", {}).get("required", False) 
                    chat_token = requirements["token"]
```

Instead of directly accessing `requirements["arkose"]`, the updated code:

1. Retrieves the data from the `text` key and parses it as JSON.
2. Checks for the presence of the `turnstile` key within the parsed JSON.
3. If `turnstile` exists, it further checks if the `required` flag is set to `True`.
4. If the `required` flag is `True`, the Arkose token is extracted from the `dx` key within `turnstile`.
5. For backward compatibility, it also checks the old `arkose` key in case it's still present.

This change ensures that the code can adapt to both the old and new API responses, allowing it to accurately detect Arkose challenges and proceed with the authentication process.

This pull request is related to this issue:
https://github.com/xtekky/gpt4free/issues/2087

### DIFF
--- a/g4f/Provider/needs_auth/OpenaiChat.py
+++ b/g4f/Provider/needs_auth/OpenaiChat.py
@@ -406,7 +406,12 @@ class OpenaiChat(AsyncGeneratorProvider, ProviderModelMixin):
                     cls._update_request_args(session)
                     await raise_for_status(response)
                     requirements = await response.json()
-                    need_arkose = requirements.get("arkose", {}).get("required")
+                    text_data = json.loads(requirements.get("text", "{}")) 
+                    need_arkose = text_data.get("turnstile", {}).get("required", False)
+                    if need_arkose:
+                        arkose_token = text_data.get("turnstile", {}).get("dx")
+                    else:
+                        need_arkose = requirements.get("arkose", {}).get("required", False) 
                     chat_token = requirements["token"]        
 
                 if need_arkose and arkose_token is None:


xtekky:
Excellent !!

Issue #2101, opened 2024-07-08
HuggingChat some model request conversationId


TheFirstNoob:
Hi, i test HuggingChat provider for avalable models and find error:
`HuggingChat: KeyError: 'conversationId'`

**Model who need this key param:**
```
HuggingFaceH4/zephyr-orpo-141b-A35b-v0.1
google/gemma-1.1-7b-it

```

Other models working fine and not request this key param

Anurag-Varma:
Verified this bug exists in HuggingFaceH4/zephyr-orpo-141b-A35b-v0.1:

![image](https://github.com/xtekky/gpt4free/assets/62068859/fad60f0d-782d-4e9c-860b-2199fbfa4ef3)


```
Using HuggingChat provider and HuggingFaceH4/zephyr-orpo-141b-A35b-v0.1 model
INFO:werkzeug:127.0.0.1 - - [11/Jul/2024 01:37:37] "GET /static/img/user.png HTTP/1.1" 304 -
ERROR:root:'conversationId'
Traceback (most recent call last):
  File "C:\Users\panur\gpt4free\g4f\gui\server\api.py", line 177, in _create_response_stream
    for chunk in ChatCompletion.create(**kwargs):
  File "C:\Users\panur\gpt4free\g4f\Provider\HuggingChat.py", line 80, in create_completion
    conversationId = response.json()['conversationId']
KeyError: 'conversationId'
INFO:werkzeug:127.0.0.1 - - [11/Jul/2024 01:37:38] "POST /backend-api/v2/conversation HTTP/1.1" 200 -
```

Anurag-Varma:
The below models are removed from [https://huggingface.co/chat/settings](https://huggingface.co/chat/settings)

```
HuggingFaceH4/zephyr-orpo-141b-A35b-v0.1
google/gemma-1.1-7b-it
```

![image](https://github.com/xtekky/gpt4free/assets/62068859/a4637560-8ab7-487b-9cd1-f67c8e279911)

Anurag-Varma:
Hi @xtekky ,

Can you check the issue, I verified it issue exists and sent a PR fix for it.

Anurag-Varma:
There is already a PR #2103

It fixes this issue, once the PR is merged, this issue can be closed

PR #2103, opened 2024-07-09
Enhancements and Updates Across Providers

kqlio67:
### Provider Updates and Improvements ###

# Blackbox
Enhanced the structure and functionality of the Blackbox class:
- Added support for models through inheritance of the ProviderModelMixin class.
- Introduced the default_model field with a default value of 'blackbox'.
- Updated the return type of the create_async_generator method to AsyncGenerator[str, None].
- Added type annotations for the image and image_name parameters, marking them as optional.
- Included the "maxTokens" field with a value of None in the data dictionary.
- Explicitly annotated the response variable as ClientResponse.
- Replaced the content method with content.iter_any() for optimized response reading.
- Implemented data cleansing from unwanted prefixes using regular expressions.
- Removed unnecessary blank lines and spaces to enhance code readability.

# ChatgptFree
Fixed the working field in the ChatgptFree class:
- Changed the working field value from False to True, indicating the ChatgptFree provider now works correctly.

# FreeGpt
Improved the structure and readability of the FreeGpt class:
- Added DOMAINS and RATE_LIMIT_ERROR_MESSAGE constants to enhance readability and avoid value duplication.
- Added type annotations for method parameters and return values.
- Renamed the `proxy` parameter to `Optional[str]` to clearly indicate it is optional.
- Updated the return type of the create_async_generator method to AsyncGenerator[str, None].
- Moved the request data construction logic to a separate static method `_build_request_data`.
- Used a ternary operator for conditional assignment of the `proxies` value in `StreamSession`.
- Renamed the `chunk` variable to `chunk_decoded` to clearly indicate its decoded nature.
- Removed unnecessary blank lines and spaces to improve code readability.

# FreeChatgpt
Improved the FreeChatgpt provider code:
- Removed unnecessary imports (ClientTimeout, raise_for_status).
- Added support for the gpt-3.5-turbo model and other models.
- Updated request headers to align with the new version.
- Modified request data format to include a system message.
- Simplified response handling using accumulated_text.
- Added exception handling for JSON parsing.
- Fixed code formatting to comply with PEP8.

# GeminiProChat
Added model support and updated URLs in GeminiProChat:
- Introduced the ProviderModelMixin class for model support.
- Added the default_model field with a default value of ''.
- Updated the URL in the Referer header field.
- Updated the URL in the Origin header field.
- Removed unnecessary comments and blank lines.

# HuggingChat
Updated the model list in the HuggingChat class:
[...]

PR #2104, opened 2024-07-10
2 model removed from HuggingChat - #2101

Anurag-Varma:
The below models are removed from [https://huggingface.co/chat/settings](https://huggingface.co/chat/settings)

```
HuggingFaceH4/zephyr-orpo-141b-A35b-v0.1
google/gemma-1.1-7b-it
```

So updated the provider by removing these models.


Anurag-Varma:
It fixes the issue #2101

Anurag-Varma:
There is already a pr #2103 which fixes this.

Closing it

PR #2110, opened 2024-07-15
Add `CheckModel` Class for Model and Provider Response Testing

codewithdark-git:
#### Description:
This pull request introduces a new class, `CheckModel`, to the `g4f` library, enabling users to test responses from different providers for a given model. The `CheckModel` class includes methods for asynchronously fetching and logging responses from available working providers.

#### Changes Made:
1. **New File**: Created `check_model.py` with the `CheckModel` class.
    - `__init__(self, model)`: Initializes with the specified model and prepares a list of working providers.
    - `get_response(self, provider)`: Asynchronously gets responses from a provider and logs results.
    - `check_model(self)`: Aggregates results from all working providers asynchronously.
2. **Modified `__init__.py`**:
    - Added `from .client.check_model import CheckModel` to make the `CheckModel` class available for import.

#### Usage:
Users can now easily check model responses by importing and running the `CheckModel` class:
```python
from g4f.client.check_model import CheckModel
import asyncio

checker = CheckModel('gpt_4_turbo')
asyncio.run(checker.check_model())
```

### DIFF
--- a/g4f/__init__.py
+++ b/g4f/__init__.py
@@ -10,6 +10,8 @@ from .cookies import get_cookies, set_cookies
 from .providers.types import ProviderType
 from .providers.base_provider import AsyncGeneratorProvider
 from .client.service import get_model_and_provider, get_last_provider
+from .client.check_model import CheckModel
+
 
 class ChatCompletion:
     @staticmethod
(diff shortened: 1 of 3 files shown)


xtekky:
The concept is very good but it can be improved because there is a lot of async issues where the provider class doesnt support it as you can see in the logs with "await not recognised"

PR #2210, opened 2024-09-02
Fix DDG: allow using conversations and improve UX for incorrect usage

0dminnimda:
The vqd-4 changes for each continuation of the converation, thus allowing stepping back or branching, but this contradicts the initial authors assumption. The logic is as such: first status update gives us first empty chat id, the successive chat queries send the new conversation id, we just got to read it and return the new one to the user.

Before that I was getting this if I were to use a conversation (DDG does  not allow more than one message without it):
```
    ...
  File "C:\Users\ASUS\AppData\Roaming\Python\Python312\site-packages\g4f\client\client.py", line 29, in iter_response
    for idx, chunk in enumerate(response):
  File "C:\Users\ASUS\AppData\Roaming\Python\Python312\site-packages\g4f\providers\base_provider.py", line 223, in create_completion
    yield loop.run_until_complete(await_callback(gen.__anext__))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python3.12.4\Lib\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\ASUS\AppData\Roaming\Python\Python312\site-packages\g4f\providers\base_provider.py", line 52, in await_callback
    return await callback()
           ^^^^^^^^^^^^^^^^
  File "C:\Python\projects\experiments\gpt4free\DDG.py", line 110, in create_async_generator
  File "C:\Users\ASUS\AppData\Roaming\Python\Python312\site-packages\g4f\requests\raise_for_status.py", line 28, in raise_for_status_async
    raise ResponseStatusError(f"Response {response.status}: {message}")
g4f.errors.ResponseStatusError: Response 400: {"action":"error","status":400,"type":"ERR_INVALID_VQD"}
```

Also it's better to make user track messages, rather than half-tracking present in here (does not track llm answer, so user have to add it to conversation messages for some reason ...) and AFAIC if a user wants to do anything useful with llm, they have to manage the messages history.

### DIFF
--- a/g4f/Provider/DDG.py
+++ b/g4f/Provider/DDG.py
@@ -8,9 +8,11 @@ import base64
 
 from .base_provider import AsyncGeneratorProvider, ProviderModelMixin
 from .helper import get_connector
+from ..errors import ResponseError
 from ..typing import AsyncResult, Messages
 from ..requests.raise_for_status import raise_for_status
 from ..providers.conversation import BaseConversation
+
 
 class DDG(AsyncGeneratorProvider, ProviderModelMixin):
     url = base64.b64decode("aHR0cHM6Ly9kdWNrZHVja2dvLmNvbS9haWNoYXQ=").decode("utf-8")
@@ -33,7 +35,7 @@ class DDG(AsyncGeneratorProvider, ProviderModelMixin):
     chat_url = base64.b64decode("aHR0cHM6Ly9kdWNrZHVja2dvLmNvbS9kdWNrY2hhdC92MS9jaGF0").decode("utf-8")
     referer = base64.b64decode("aHR0cHM6Ly9kdWNrZHVja2dvLmNvbS8=").decode("utf-8")
     origin = base64.b64decode("aHR0cHM6Ly9kdWNrZHVja2dvLmNvbQ==").decode("utf-8")
-    
+
     user_agent = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36'
     headers = {
         'User-Agent': user_agent,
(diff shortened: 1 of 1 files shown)


g4f/Provider/DDG.py
```diff
@@ -74,32 +76,38 @@ async def create_async_generator(
...
+                        raise ResponseError(
+                            f"Status {response.status}: Conversation does not exactly match the messages or its id (vqd-4) is invalid"
+                        )
                 await raise_for_status(response)
                 if return_conversation:
-                    yield Conversation(vqd_4, messages)
-                
+                    vqd_4_new = response.headers.get("x-vqd-4")
+                    if vqd_4_new is not None:
+                        yield Conversation(vqd_4_new, messages)
```
0dminnimda, line 109: Should I raise an exception if we could not get the conversation id, but the user requested it? Currently it's not a silent error, the user can detect an absence of the Converation chunk.

TheFirstNoob:
If i correclty remeber ddg site give 50 messages Converation and after need press "Fire" (clear button).
Maybe need add checking for `>49` like

kqlio67 (approved):

kqlio67:
Thank you for these valuable improvements and the detailed explanation about vqd-4 behavior. Your changes significantly enhance error handling and conversation management. I appreciate your efforts in improving the `DDG` provider. These enhancements make the provider more robust and efficient.

TheFirstNoob:
DDG: ValueError: More than one message requires using conversation.

This happening after first message sended or if some "bots" use conversation saving well we have error to

0dminnimda:
> DDG: ValueError: More than one message requires using conversation.

Yes, otherwise you'll get an error I referenced in the ticket description.

0dminnimda:
> If i correclty remeber ddg site give 50 messages Converation and after need press "Fire" (clear button). Maybe need add checking for `>49` like

api doesn't seem to have that limitation, I tested it handling 160 short prompts and still recalling information in the first prompt

0dminnimda:
> DDG: ValueError: More than one message requires using conversation.
> 
> This happening after first message sended or if some "bots" use conversation saving well we have error to

yes, if you don't use conversation you'll get an error from ddg, you just have to get the conversation, like this:

```
        if isinstance(chunk, BaseConversation):
            self.conversation = chunk
```

and pass it back to the next completion like:

```
self.provider.create_completion(
    ...
   conversation=self.conversation
)
```

Isn't that the whole "secure llm" idea, that you can continue only with a token, because continuation can contain private information from the chat?

0dminnimda:
It would be good to reopen this PR

PR #2219, opened 2024-09-11
feat(GeminiPro): add support for system prompt

monosans:
See <https://ai.google.dev/gemini-api/docs/system-instructions?lang=rest>


monosans:
bump

PR #2310, opened 2024-10-30
Compatibility for Api requests from continue.dev throught openwebui

hqnicolas:
this pull request is just an example
it fixes the api compatibility for continue.dev to be controlled by openwebui
Also disabled Version check (it need to be fixed and i'm not the gui)

please edit this pull request with a fix for the version check online

### DIFF
--- a/g4f/models.py
+++ b/g4f/models.py
@@ -120,52 +120,52 @@ default = Model(
 ### OpenAI ###
 # gpt-3
 gpt_3 = Model(
-    name          = 'gpt-3',
+    name          = 'gpt-3:latest',
     base_provider = 'OpenAI',
     best_provider = NexraChatGPT
 )
 
 # gpt-3.5
 gpt_35_turbo = Model(
-    name          = 'gpt-3.5-turbo',
+    name          = 'gpt-3.5-turbo:latest',
     base_provider = 'OpenAI',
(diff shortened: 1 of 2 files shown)


kqlio67:
Hi @hqnicolas,

As a fellow g4f user, I have concerns about the proposed changes adding ":latest" suffix to model names in models.py.

While I appreciate your efforts to improve continue.dev compatibility, I believe this change could create significant issues:

**1. API Compatibility Issues:**
The standard model naming convention used by OpenAI and other providers follows patterns like "gpt-4". Adding ":latest" breaks this convention and could cause API failures.

**2. Breaking Changes for Existing Users:**
Many of us have built projects using the current model names like "gpt-3.5-turbo". Forcing everyone to update their code to use "gpt-3.5-turbo:latest" would be disruptive and potentially break many existing implementations.

**3. Integration Problems:**
APIs expect specific model names - they won't recognize versions with ":latest" suffix. This means our calls could fail when providers try to validate model names.

**Instead of this approach, perhaps we could:**
- Keep standard model names for API compatibility
- Handle versioning in a way that doesn't modify model identifiers
- Look for a solution that maintains backward compatibility

I'm not a project maintainer, but as someone who relies on g4f, I think we need an approach that won't force users to significantly modify their existing code.

Issue #2274, opened 2024-10-14
Can't use module logging, when I use module g4f


Shwumper:
When I import module logging, I can't use it. It doesn't work. When I remove module g4f, I can use module logging.
![image](https://github.com/user-attachments/assets/e166e01b-672c-416a-8777-dc62e46ee7f1)
![image](https://github.com/user-attachments/assets/dca84b1d-ce97-43af-af5d-d7be676e83c3)
![image](https://github.com/user-attachments/assets/fc446013-6133-48ff-b760-0d5a682e4c05)

foxfire52:
You should post the error message and your full python imports.

Shwumper:
Hello, I don't have any error messages, here are full imports:
![image](https://github.com/user-attachments/assets/32920fbf-c5e3-4792-a700-b2d0dbebdcb1)

foxfire52:
if you use basicConfig() try setting a higher logging level like **level=logging.DEBUG**  and check for errors. I will try replicating this later.

Shwumper:
Thank you

foxfire52:
The problem is basicConfig(). As workaround you can add parameter **force=True** to basicConfig() to fix it. I will open another issue.

Issue #2173, opened 2024-08-22
Disable g4f logging completely


scuty2000:
**Bug description**
In my script I have my customized logging, but whenever I use it it prints 2 times (one from my logger, one from g4f logger).
How can I turn off the logger inside the library? Already tried a bunch of stuff with no results.

P.S. Are you using the root logger maybe? If that is the case, please use it with the module name

ex.
1. Create a new logger in new class
2. Set logging level to DEBUG
3. Log something
4. Enjoy duplicated output

**Screenshots**
![image](https://github.com/user-attachments/assets/c24a1b65-b513-4449-8842-870552289de0)

**Environment**
- python version: 3.11
- location ( are you in a cloudfare flagged country ) ? nope

scuty2000:
With futher inverstigation, disabling all loggers except the one I create, still have the same problem. Don't know what you did with the logger. It only occurs after I import the g4f library.

This is the code I used to disable all loggers except mine.

```python
    logger_allowlist = [
        "dataset_generation",
        # "datasets",
        # "huggingface_hub",
    ]


    for log_name, log_obj in logging.Logger.manager.loggerDict.items():
        if not any([log_name.startswith(allow) for allow in logger_allowlist]):
            log_obj.disabled = True
        if log_obj.__class__ != PlaceHolder and log_obj.disabled == False:
            print(log_name)
```

PR #2347, opened 2024-11-12
Improve logging implementation

foxfire52:
Current logging implementation needs improvement.

**Issues**
1. Code should avoid usage of **basicConfig()** to set up logging since subsequent calls by library users might fail silently.
The function `basicConfig()` sets up the **root** Logger with a **StreamHandler**. 
Since `basicConfig()` function ["**does nothing** if the root logger already has handlers configured"](https://docs.python.org/3/library/logging.html#logging.basicConfig), after **g4f code** calls it **library users** will always experience silent failure when calling it again (issue #2274).
Note that configuration of the **root logger** should be avoided by libraries (see [Configuring Logging for a Library - Python Docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)).
 
3. Logging functions are called from **root logger**. This is not a recommended practice and can interfere with custom logging setups  (log duplication issue #2173) and third-party libraries.

4. There is leftover unused code in [g4f/api/_logging.py](https://github.com/xtekky/gpt4free/blob/main/g4f/api/_logging.py), which sets up handlers with **basicConfig()**. If called it might fail silently, since there's another `basicConfig()` call in  [g4f/Provider/you/har_file.py#L14](https://github.com/xtekky/gpt4free/blob/main/g4f/Provider/you/har_file.py#L14)  (see issue **1**).
Module **Loguru** is only imported here, so it's unused as well.

**Proposed solutions**
1. Replace call to `basicConfig()` with **setLevel()**
2. Create a **named Logger** for the library.
3. **Remove** or **comment out** unused code and remove **Loguru** dependency.

**Info:** Current code sets logging error level to **ERROR**.


hlohaus:
Thank you

PR #2351, opened 2024-11-13
feat: Convert HTML entities back to their corresponding characters in Ai4Chat response

Simatwa:
A response like `I think you&#39;re amazing!` will be `I think you're amazing! `


xtekky:
Ai4Chat is now deprecated

PR #2353, opened 2024-11-14
FIX: fix the url in the markdown output (md compliance), and the chunk/ImageResponse to str !

manatlan:
add fix_url ( replace ' ' by '+' (to be markdown compliant))

### DIFF
--- a/g4f/image.py
+++ b/g4f/image.py
@@ -23,6 +23,11 @@ EXTENSIONS_MAP: dict[str, str] = {
     "image/webp": "webp",
 }
 
+def fix_url(url:str) -> str:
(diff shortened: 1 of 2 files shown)


manatlan:
url should not contains spaces ... if it's the case, some markdown tools (marked/js, or vscode) can't interpret correctly.

So replacing space with "+" fix it defenitivly

manatlan:
add another fix on "g4f/providers/base_provider.py"
to be able to join string correctly. (when chunk is an ImageResponse)

manatlan:
add a fix_url for solo image response too !

All tested and approved, by me 
now it works

PR #2422, opened 2024-11-25
upload image as data uri for conversation api

vuthaihoc:


hlohaus:
Thank you. Does the image have to be in base64 format? Should we also support URLs?

hlohaus:
It doesn't need to be pillow image for the providers.
I added only a is_data_uri_an_image check:
https://github.com/xtekky/gpt4free/pull/2423
Then it works also without pillow.

hlohaus:
I integrated the image-sending functionality with the Chat Completion API for seamless interaction.

Issue #2349, opened 2024-11-13
Providing a docker/image for "linux/arm/v7" (raspberry pi) ... could be marvelous !


manatlan:
**Is your feature request related to a problem? Please describe.**
no

**Describe the solution you'd like**
Providing a docker/image for "linux/arm/v7" (raspberry pi) ... could be marvelous !

**Describe alternatives you've considered**
I tried a docker-compose of the github repo, with no luck (compilation trouble)
But I'm really not an expert ... perhaps an expert could help ?

**Additional context**
it could be marvelous to have a docker running on a simple raspberry pi !

manatlan:
Do you known what components is problematic for armv7 ?
Perhaps I could help to fix them.

foxfire52:
@manatlan The project uses **seleniarm/node-chromium** as base image. I think the selenium mainteners don't provide an **armv7** image. You could ask here https://github.com/SeleniumHQ/docker-selenium

foxfire52:
@manatlan Try replacing the first line in **Dockerfile**   https://github.com/xtekky/gpt4free/blob/0d8e4ffa2c0706b0381f53c3985d04255b7170f5/docker/Dockerfile#L1-L5 
with
`FROM seleniarm/node-chromium:4.18.1-20240326`
Then run `docker compose up --build` on your **arm** device.

It's an older selenium docker image with **armv7** support.

manatlan:
I tried ... but got this problem 
```
80.45             building '_cffi_backend' extension
80.45             creating build/temp.linux-armv7l-cpython-311/src/c
80.45             arm-linux-gnueabihf-gcc -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC -DFFI_BUILDING=1 -DUSE__THREAD -DHAVE_SYNC_SYNCHRONIZE -I/usr/include/ffi -I/usr/include/libffi -I/usr/include/python3.11 -c src/c/_cffi_backend.c -o build/temp.linux-armv7l-cpython-311/src/c/_cffi_backend.o
80.45             src/c/_cffi_backend.c:15:10: fatal error: ffi.h: No such file or directory
80.45                15 | #include <ffi.h>
80.45                   |          ^~~~~~~
80.45             compilation terminated.
80.45             error: command '/usr/bin/arm-linux-gnueabihf-gcc' failed with exit code 1
80.45             [end of output]
[...]

PR #2427, opened 2024-11-25
Fix Armv7 docker image

foxfire52:
#2349 @hlohaus

Current **armv7** image fails build of critical components. My image should work.

You can test a local build on **ubuntu x64** with these commands from `gpt4free` folder
```bash
#qemu setup
docker run --privileged --rm tonistiigi/binfmt --install all

docker buildx build --platform linux/arm/v7 . -f docker/Dockerfile-slim 2>&1 | tee docker_build.log
```
Check the log file for **ERROR**s.

I removed `duckduckgo-search` because `primp` can't compile. If rust is installed it fails building `boring-sys`. These bugs could be related  [link1](https://www.github.com/cloudflare/boring/issues/283) [link2](https://www.github.com/cloudflare/boring/issues/242). I recommend waiting for the [PR](https://www.github.com/cloudflare/boring/pull/223).

If you want to test the `boring-sys` bug add these lines at **line 50** and build again.

```
USER root
RUN apt install -qqy curl

USER $G4F_USER_ID
ENV PATH "${HOME}/.local/bin:${HOME}/.cargo/bin:${PATH}"
RUN curl https://sh.rustup.rs/ -sSf | bash -s -- -y
RUN pip install primp duckduckgo-search>=5.0
```

### DIFF
--- a/docker/Dockerfile-slim
+++ b/docker/Dockerfile-slim
@@ -26,6 +26,7 @@ USER $G4F_USER_ID
 WORKDIR $G4F_DIR
 
 ENV HOME /home/$G4F_USER
+ENV PATH "${HOME}/.local/bin:${PATH}"
 
 # Create app dir and copy the project's requirements file into it
 RUN mkdir -p $G4F_DIR
--- a/docker/Dockerfile-slim
+++ b/docker/Dockerfile-slim
@@ -3,6 +3,7 @@ FROM python:slim-bookworm
 ARG G4F_VERSION
 ARG G4F_USER=g4f
 ARG G4F_USER_ID=1000
+ARG PYDANTIC_VERSION=1.8.1
 
 ENV G4F_VERSION $G4F_VERSION
 ENV G4F_USER $G4F_USER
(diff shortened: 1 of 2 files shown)


hlohaus:
I added a Dockerfile-armv7 in my pull request. I think that is better.

foxfire52:
Instead of splitting the Dockerfile we could do something like
```
RUN if ["$TARGETPLATFORM" = "linux/arm/v7"]; then \
...
    pip install --no-cache-dir duckduckgo-search>=5.0
...
```

I haven't tested your file yet, does it compile duckduckgo-search correctly? We don't need to install rust on amd64, right?

foxfire52:
> We don’t need any build tools for amd64/arm64. It’s better to avoid installing unnecessary tools, as that eliminates the need to uninstall them later.

Good

> For web search functionality, the GUI only requires DuckDuckGo Search. The API, however, does not depend on it.
Understood. My only concern was avoiding Dockerfile duplication, but a different Dockerfile might be better.

deedy5:
`duckduckgo_search` v6.3.7 is compatible with `armv7`

hlohaus:
I added duckduckgo_search again.

foxfire52:
@deedy5 Thank you for adding `primp` armv7 support.

PR #2434, opened 2024-11-26
Fix docker armv7 cffi pillow

foxfire52:
#2349 #2427
This should fix missing headers for **cffi** and **pillow** libraries.


hlohaus:
Thanks, I've merged it.

PR #2436, opened 2024-11-27
Fix docker armv7 cryptography module build

foxfire52:
#2349 
This should fix **cryptography** module build error.
We can't clean up rust with `rustup self uninstall -y` because there's a bug `Value too large` [link](https://www.github.com/rust-lang/cargo/issues/8719), but we can remove `.rustup` folder.


hlohaus:
g4f Copilot says: ready to be merged 

Thank you

Issue #2433, opened 2024-11-26
Issues with Blackbox Provider


DeveleplorAbdullahH:
There are so many issues with Blackbox provider. It can't read the system prompt and gpt-4o model can not answer. Maybe, there are issues with the servers.

Linden10:
I asked this same question and the answer is here:
https://github.com/xtekky/gpt4free/issues/2360#issuecomment-2495509264
> @Linden10 Hi, Blackbox is sometimes overloaded and because of this the dialogue, instructions or system settings are lost. For this provider such data losses are quite normal already. That's why I didn't add this provider to the bug list.

I tested this manually with Blackbox in g4f gui, sending messages one at a time and testing the system prompt and found out that Blackbox randomly breaks, ignoring the system prompt and such after responding due to the reasons above.

Yup! Nothing that can be done about it sadly...

DeveleplorAbdullahH:
And there is an issue with Copilot Provider. I wrote a long system prompt to Copilot provider and when I ask something to Copilot provider it says: "RuntimeError: Error: {'event': 'error', 'id': '0', 'errorCode': 'text-too-long'}"

Issue #2393, opened 2024-11-21
ChatgptES has an english version called chatgpt.co.uk and missing system prompt


Linden10:
I've been using GPT4Free with an python project called DazedMTL to translate Japanese games into English!

Whenever I try to translate lines from Japanese to English with ChatGptEs, the provider never responds in English but in Spanish and I had to disable it as it was inconvenient and never translating any of the lines I provided.

I only once then decided to visit the site and discovered on the top-right, in a menu that it has an English version of the site called chatgpt.co.uk! To test if it works with the api, I manually modified the Chatgptes.py module and replaced the urls with the site and it works! Everything's in english which is great!

If you can, can you add this (or switch ChatGptEs if you want to the English version) to G4F? Thank you!

...Also if it's possible, can you make it so it can accept a system prompt so that it doesn't respond randomly like this:

> It looks like you've shared a text that involves a conversation between characters, focusing on themes of tension and relaxation in a somewhat intimate setting. The characters are discussing how to alleviate nervousness and create a more comfortable atmosphere.
> 
> If you have any specific questions about the text or need help with something related to it, feel free to ask! I'm here to help!
> 

This is a response from the english verison of the provider but it was always like this in Spanish as well, never translating or responding in json format, ignoring my prompt I have setup.

Every other provider (using gpt-4o providers) always read the json formatted lines and responds with the translated json text thanks to the prompt explaining what to do with them (well...until they stop and give the generic "I'm sorry, but I can't assist with that." response whenever the text gets spicy ha)...all except for this provider weirdly!

I tested using the prompt directly on chatgptes/chatgpt.co.uk and it works on there suggesting...the system prompt isn't setup in the code based on what I looked at.

If it's possible to include the prompt as a message, please do add it as a feature too! ^^

Linden10:
Now I'm starting to notice that another API Provider, blackbox, is doing the same thing in terms of ignoring my system prompt and just responding in Japanese whenever it reads the json japanese lines I send for translation...

I also noticed this happening to another user that uses DazedMTL, the same thing happening to them whenever Blackbox responds. I'm wondering if the system prompt is being sent or ignored at times...

Or is there like a limit to how big the system prompt is supposed to be for Blackbox?
If there's a token limit, maybe something like that is causing the problem I wonder...

If there's a simple solution to this I would love to hear it.
Anyways here's a typical response from Blackbox when sending a translation:
```
[...]

Issue #2360, opened 2024-11-15
[Bug-List/Requests] New/Missing models, providers errors, Minor improvements


TheFirstNoob:
### Actual for v0.3.8.1
List bugs and requests for g4f in issues list
This list will allow you to keep current issues and requests in case some of the topics were closed over time or lost.
**Please mark me if something is already fixed or I forgot to add a confirmed error. I will add/remove it.**

### REQUESTS: 
1. `Airforce` provider have more models and requested to add/restore `models.py` list (Need revert file to 0.3.3.4 version @kqlio67  version) + need split like `api.airforce` and `llmplayground` provider.
- For **Airforce** need use v0.3.3.4 version file (More chat-models, more stability, more faster response)
- For **LLM** need use actual version file (Actual LLM response but small stable + small models)

All this models work correctly on v0.3.3.4 version file!
- `gpt-3.5-turbo`
- `claude-3-haiku`
- `claude-3.5-sonnet`
- `gemini-flash`
- `gemini-pro`
- `llama-3.1-405b`
- `llama-3.2-90b`
- `mixtral-8x22b` https://github.com/xtekky/gpt4free/issues/2359
- `yi-34b`

2. Add Mistral large model with vision if possible https://github.com/xtekky/gpt4free/issues/2386
    - I think is new *Le Chat* model: https://chat.mistral.ai/chat
3. Add basic Auth/Managment user logic for WebUi: https://github.com/xtekky/gpt4free/issues/2407

### BUGS:
1. `PerplexityLabs` for `sonar-online` model response: `Unknown error`
2. `PizzaGPT` provider return `empty message` https://github.com/xtekky/gpt4free/issues/2392
3. `MagickPen` provider return:  `Unable to extract all the necessary data from the JavaScript file`
4. `Prodia` provider request cookies cause calling bing provider: `Error reading cookies from edge for .bing.com: ` (Only if try to use direct model call)

AndreySupr:
request ended with status code 404

I have the same problem. Tell me, did you manage to fix it?
Version: 0.3.3.3

hlohaus:
Hey Andrey, which provider did you use? Maybe try it with a different one.

TheFirstNoob:
@hlohaus Hey.
1. I think user above tell about RubiksAI before fix we have 404 error. Other providers i didnt remember with this error.
2. I always try to check repo changes and update this thread for provider actual list bugs and requests. Please can you provide that i miss remove after fix/updates?

[...]

PR #2437, opened 2024-11-27
feat(g4f): Major provider updates and new model support

kqlio67:
**Core changes:**
- **Added new models:** `o1-preview, o1-mini, llama-3.2-90b, codellama-34b, mixtral-7b, mistral-tiny, claude-3.5-haiku, qwen-2.5-72b, flux-dev, dalle-3, recraft-v3`
- **Added new providers:** `Copilot, LLMPlayground, AmigoChat`
- **Added new base providers:** `Gryphe, databricks, anthracite-org, ai21, llmplayground.net`

**Provider improvements:**
- Enhanced `DDG` and `Copilot` with better conversation handling
- Fixed system prompts in `ChatGptEs` and `Blackbox` (#2393), (https://github.com/xtekky/gpt4free/issues/2433#issue-2695811324)
- Improved message processing in `Airforce` (#2360)
- Updated `Liaobots` and `DeepInfra` with latest models
- Added `Qwen2.5` and `Nemotron` support to multiple providers

**Maintenance:**
- Moved `PollinationsAI` to `needs_auth/`
- Moved `AIUncensored` to `not_working/`
- Removed deprecated `Allyfy` provider
- Updated model mappings and provider configurations

**Breaking changes:**
- Removed several outdated models from `Liaobots`


TheFirstNoob:
Hi! I still think `airforce` and `llm` providers should be divided into two providers, and not replace each other. Based on tests from version 0.3.3.4 and any subsequent ones, if you use llm settings, then there are much fewer working providers.
If you use airforce on version 0.3.3.4, then there are much more working models, although they are not specified there.

Also, `Chatgpt4Online` and `FreeNetfly` are not removed from the `model.py`. These providers are removed, but we still call them in this file.

TheFirstNoob:
@kqlio67 Thanks for answer! Let me explain in more detail.
The file version from version 0.3.3.4 works with models that are not listed on the site itself, they were listed there earlier, but the functionality works and quite quickly. This list of models is listed above in the bug list and they must be specified in `models.py`. At the same time, the LLM version, that is, everything after 0.3.3.4, works extremely strangely. On `3.5-turbo` and those models that are listed in the bug list, they give me an error that "such a model is not supported" (since we are checking for models that are listed in `/models/v1`. And all subsequent models, when requested, freeze for me and endlessly wait for a response.

It is for this reason that I suggest separating these two file versions, since there will be greater availability of old and new models. Or a different approach is required.

PR #2469, opened 2024-12-08
Updating provider documentation and small fixes in providers

kqlio67:


hlohaus:
Thank you. I have updated the model list in models.py too.
I added a unittest to check the model s are in the provider's model list.

TheFirstNoob:
@kqlio67 Hi.
I have implemented a similar idea with "Continue" in my test form as "Send text: Continue + last line of text". This is not very stable because the model can either correctly continue or, on the contrary, repeat some part of the answer from the previous text, but as a fast function it works quite well.
I think something similar can be implemented.

hlohaus:
Thank you

PR #2482, opened 2024-12-14
Add Clarity Token to Copilot,

hlohaus:
Ignore disabled providers in gui,
Read cookie files after upload

### DIFF
--- a/g4f/Provider/Airforce.py
+++ b/g4f/Provider/Airforce.py
@@ -32,8 +32,7 @@ class Airforce(AsyncGeneratorProvider, ProviderModelMixin):
     api_endpoint_imagine2 = "https://api.airforce/imagine2"
 
     working = True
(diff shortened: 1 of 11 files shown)


Arken1xs:
@hlohaus can I ask one question, please? Why did you changed the Airforce aittribute "supports_stream" to False, if it **does support** the stream? Airforce (as it seems, I'm not completely sure) does not support **non-stream**, but **stream works well**.

hlohaus:
@Arken1xs It doesn't yield the chunks to the client. All chunk will append to a buffer and will be returned at the end. It isn't streaming.

Arken1xs:
> @Arken1xs It doesn't yield the chunks to the client. All chunk will append to a buffer and will be returned at the end. It isn't streaming.

Yeah, but now, after update, Airforce does not work at all.

PR #2516, opened 2024-12-28
fix: Update deprecated DuckDuckGo search backend from 'api' to 'auto'

kqlio67:
Fixes UserWarning: 'api' backend is deprecated, using backend='auto'
- Updated default backend parameter from 'api' to 'auto' in search function
- Aligns with latest duckduckgo-search library recommendations


hlohaus:
Thank you, @kqlio67

PR #2540, opened 2025-01-05
Add Edge as Browser for nodriver

hlohaus:
Fix for RetryProviders doesn't retry
Add retry and continue for DuckDuckGo provider
Add cache for Cloudflare provider
Add cache for prompts on gui home
Add scroll to bottom checkbox in gui
Improve prompts on home gui
Fix response content type in api for files


HAK3R4LIFE:
Umm Firefox is better ☝️🤓

Sent from Proton Mail Android

-------- Original Message --------

PR #2549, opened 2025-01-08
Add Jmuz provider

evqsx:


hlohaus:
@evqsx Does it having a login url? It use it in the GUI: Example:
    login_url = "https://platform.openai.com/settings/organization/api-keys"

evqsx:
@hlohaus Hi, it doesn't have a login GUI, it's just a simple API. The owner is still building the API Keys system, so just the free key is currently available. Not sure when the full thing will be completed.

hlohaus:
@evqsx Let's run the unitests

PR #2570, opened 2025-01-14
Provider Updates and Fixes

kqlio67:


TheFirstNoob:
@kqlio67 Hi!
We have minor issues for alias model name for some provider.


1. hf_space/Qwen_QVQ_72B.py

Need create new alias name. Not need use qwq-32 cause its difference models


2. /needs_auth/GlhfChat.py

Most name for alias not needed postfix `Instruct` or other cause we have this models in other provider alias


This should be more effective and more compatibility for model calling.

hlohaus:
Thank you!

PR #2594, opened 2025-01-24
Update provider parameters, check for valid provider

hlohaus:
Fix reading model list in GeminiPro
Fix  check content-type in OpenaiAPI

### DIFF
--- a/g4f/Provider/needs_auth/GeminiPro.py
+++ b/g4f/Provider/needs_auth/GeminiPro.py
@@ -23,6 +23,7 @@ class GeminiPro(AsyncGeneratorProvider, ProviderModelMixin):
 
     working = True
     supports_message_history = True
+    supports_system_message = True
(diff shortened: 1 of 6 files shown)


TheFirstNoob:
@hlohaus Hi. A bit off topic, but I'll write it for your convenience. In order not to forget to update the link to the release version every time, you can use this one: https://github.com/xtekky/gpt4free/releases/latest
When a new version is released, it is usually always marked as the latest. This is a more universal solution.

hlohaus:
> @hlohaus Hi. A bit off topic, but I'll write it for your convenience. In order not to forget to update the link to the release version every time, you can use this one: https://github.com/xtekky/gpt4free/releases/latest When a new version is released, it is usually always marked as the latest. This is a more universal solution.

@TheFirstNoob 
I'm not currently creating a Windows package with every version update. 😉

PR #2679, opened 2025-02-03
Updated some providers, added new providers and added new models

kqlio67:


TheFirstNoob:
@kqlio67 Hi, can you please check DeepInfraChat provider for Not Auth error (On free time). I know you fix this on some PR before but on latest version i still have this error. Problem on my region or?
Thanks <3

TheFirstNoob:
And some minor moments:

1. DeepInfraChat now have some more deepseek models:
image

2. GPRO provider can be disable cause 429 quota limit.
I think I won't offer to return it anymore even if it works, because the provider's quota runs out in a couple of days. Maybe mark it with a comment so as not to move it around folders many times?

PR #2683, opened 2025-02-04
refactor: use dependency injection for RequestConfig

sobelmangentz:
Use dependency injection for RequestConfig so that other can overwrite it.


sobelmangentz:
@hlohaus Can you please review this PR?

PR #2717, opened 2025-02-14
Add o3-mini and Llama 3.3 70B Instruct Turbo to DDG (DuckDuckGo) provider

Arken1xs:
This little change adds one new model - **o3-mini** - and replaces previously used Llama 3.1 70B with **Llama 3.3 70B Instruct** to the DDG provider.


hlohaus:
Thank you

PR #2734, opened 2025-02-21
Improve tools support in OpenaiTemplate and GeminiPro

hlohaus:
Update models in DDG, PerplexityLabs, Gemini
Fix issues with curl_cffi in new versions


TheFirstNoob:
Hey, for DDG maybe use new alias? `"mixtral-8x7b":` to `mixtral-small`?
This will avoid confusion for providers who still use the old 8x7 model.

PR #2816, opened 2025-03-21
fix interference api usage docs with open ai

alwaz-shahid:


hlohaus:
What are the fix?
@alwaz-shahid 
You only changed the messages, the api_key and streaming.

Issue #2806, opened 2025-03-16
Update Reka provider url


TheFirstNoob:
Now url is: https://space.reka.ai/ (redirected from chat.reka.ai)
Now support thinking mode

PR #2823, opened 2025-03-22
Refactor Reka provider class: Update URL and url handling

nenas1ya:
- Change old URL to new  #2806 
- Updated error messages/body content to reference cls.domain/cls.url instead of hardcoded values for better maintainability.


hlohaus:
Thank you

PR #2883, opened 2025-04-02
Fix Slim Docker instructions

voronkovich:
I guess all Python commands should be run inside the container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,9 +119,7 @@ docker run \
   -v ${PWD}/har_and_cookies:/app/har_and_cookies \
   -v ${PWD}/generated_images:/app/generated_images \
   hlohaus789/g4f:latest-slim \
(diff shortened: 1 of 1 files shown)


hlohaus:
Hey Voronkovich, they're running inside the container for me.  Using an old Docker version? @voronkovich

voronkovich:
@hlohaus, I thinks it doesn't depend on the Docker version. The `&&` keyword is evaluated by the shell (bash, zsh).

For example:

```sh
docker run --rm hello-world && whoami
```

The `whoami` command is executed on the host machine, not in the container.

Try this command:

```sh
docker run --rm hlohaus789/g4f:latest-slim whoami && whoami
```

What do you see?

hlohaus:
@voronkovich , I am utilizing a Linux system.  The issue may originate from your operating system.

PR #2910, opened 2025-04-15
fix(ImageGenerationConfig): add size="WxH" support

BuT9l3b:
- Parse `size` string (e.g., "256x256") to auto-set `width` and `height`
- Keep explicit `width`/`height` priority if both provided

Fixes xtekky#2902


hlohaus:
Thank you!

PR #2942, opened 2025-04-21
fix: serialize None as empty string

GravityTwoG:
**Problem**:

content with value None is serialized as `"None"` string in response when using stream mode:
image


hlohaus:
@GravityTwoG 

Thank you 🫡

Issue #2941, opened 2025-04-21
OpenAI compatible provider list returns Error 500


theDaRkMaN1984:
Hey!

If you for example check the provider list under http://localhost:1337/v1/providers
everything is fine. If you try to check the modellist of a provider for example: http://localhost:1337/v1/providers/DeepInfraChat

it returns an error 500 Internal Server Error.
Sorry that i couldn't check which change kills it.

```
INFO:     172.17.0.1:49560 - "GET /v1/providers/DeepInfraChat HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
  + Exception Group Traceback (most recent call last):
  |   File "/home/g4f/.local/lib/python3.13/site-packages/starlette/_utils.py", line 76, in collapse_excgroups
  |     yield
  |   File "/home/g4f/.local/lib/python3.13/site-packages/starlette/middleware/base.py", line 177, in __call__
  |     async with anyio.create_task_group() as task_group:
  |                ~~~~~~~~~~~~~~~~~~~~~~~^^
  |   File "/home/g4f/.local/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 772, in __aexit__
  |     raise BaseExceptionGroup(
  |         "unhandled errors in a TaskGroup", self._exceptions
  |     ) from None
  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "/home/g4f/.local/lib/python3.13/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    |     result = await app(  # type: ignore[func-returns-value]
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         self.scope, self.receive, self.send
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |     )
    |     ^
    |   File "/home/g4f/.local/lib/python3.13/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    |     return await self.app(scope, receive, send)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/home/g4f/.local/lib/python3.13/site-packages/fastapi/applications.py", line 1054, in __call__
    |     await super().__call__(scope, receive, send)
[...]

PR #2943, opened 2025-04-21
Fix: Resolve NameError in provider info endpoint (#2941)

BuT9l3b:
Corrected access to ProviderUtils by using its full path g4f.Provider.ProviderUtils.


hlohaus:
Hey @BuT9l3b, we import `Provider`, 
better then `g4f.Provider`.

hlohaus:
Thank you! 😁 @BuT9l3b

PR #2973, opened 2025-05-01
Update Qwen_Qwen_3.py

SreejiBR:
This PR addresses an issue where some text type updates from the process_generating stream were not being yielded, resulting in missing output content, particularly content that followed the reasoning section.

**Root Cause**
update[2] blocks with type: "text" were not handled unless embedded in certain paths. As a result, standalone or final text entries were silently ignored during streaming.

**Solution**
Added explicit handling for type: "text" updates.


hlohaus:
Thank you 👍

PR #2988, opened 2025-05-14
Remove deprecated requirements files: requirements-min.txt and requirements-slim.txt

skewdah:


hlohaus:
They are not deprecated and your commit is broken

PR #3026, opened 2025-06-06
fix: ignore /message/metadata/refresh_key_info

GravityTwoG:
OpenaiAccount Completions API adds ids from /message/metadata/refresh_key_info event to the response if image generation is used

Screenshot 2025-06-06 at 14 57 59


hlohaus:
Thank you 👍

REDMEE1:
[``](I’ll)

PR #3117, opened 2025-08-07
Feat/provider Nvidia

lorriscript:
Added the Nvidia NIM APIs Provider


hlohaus:
Thank you! @lorriscript

Issue #3172, opened 2025-09-08
Add kimi-k2-0905 and qwen3-max


vinogradgarden:
**Is your feature request related to a problem? Please describe.**
No, it's not.

**Describe the solution you'd like**
Add new Moonshot and Alibaba flagship models.

**Additional context**
Moonshot and Alibaba have just released new models.

PR #3173, opened 2025-09-08
feat(qwen): add support for qwen3-max-preview model

restartxx:
This adds the 'qwen3-max-preview' model to the list of supported models for the chat.qwen.ai provider.

Partially addresses #3172.


hlohaus:
Thank you @restartxx

PR #3224, opened 2025-10-30
Update HuggingFaceMedia.py

kirill670:
Fixed resolution error with veo and sora2.


hlohaus:
HuggingFaceMedia.py doesn't has veo or sora2?

PR #3237, opened 2025-11-07
Improve auth cookie detection in Copilot provider

3mora2:
Adds a check for 'auth0' in cookies when authentication is required, ensuring the loop breaks appropriately once the relevant cookie is present.


hlohaus:
Thank you!

PR #3276, opened 2025-12-06
fix g4f/Provider/Qwen.py, g4f/Provider/needs_auth/OpenaiChat.py

3mora2:
use wss in OpenaiChat to get task media


hlohaus:
@3mora2 thank you

PR #3319, opened 2026-01-10
Yupp

hlohaus:


3mora2:
can use this version
https://github.com/zinzied/cloudscraper
it support async and use aiohttp

PR #3321, opened 2026-01-12
Fix Qwen provider implementation

moisestohias:
This PR includes fixes for the Qwen provider:
- Re-enable cookie generation for authentication
- Update User-Agent header to latest version
- Fix Authorization header to only be added when token is present
- Add X-Requested-With and X-Source headers for proper API communication
- Force stream and incremental_output to True for consistent streaming behavior
- Set parent_id to None to prevent parent-child relationship issues
- Add conditional thinking functionality with proper feature_config when enable_thinking is True
- Improve header management for better compatibility with Qwen API


hlohaus:
@moisestohias 

Hello, why you are removing the conversation support? (parentId)

Thank you

PR #3349, opened 2026-02-03
Fix: publish main Docker image as multi-arch (amd64 + arm64)

calmintp:
The main Docker image (hlohaus789/g4f:latest) is currently published as amd64-only due to the CI configuration.

Although the Dockerfile and base image already support arm64, the build workflow explicitly restricts the platform to linux/amd64.

This PR enables multi-arch builds (amd64 + arm64) for the main image, fixing Docker pulls on Apple Silicon and ARM machines.

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -290,7 +290,7 @@ jobs:
         with:
           context: .
           file: docker/Dockerfile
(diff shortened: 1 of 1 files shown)


hlohaus:
The base Dockerfile/image doesn't support this architecture. This build change doesn't fix it. Please test your changes.

PR #3352, opened 2026-02-07
feat: add tool_call emulation for OpenAI API

mvclaudianobj:
Avoid forcing PollinationsAI when tools are present, and add an opt-in tool_emulation mode (or G4F_TOOL_EMULATION=1) to emit OpenAI-compatible tool_calls for providers that ignore tools.

### DIFF
--- a/etc/unittest/__main__.py
+++ b/etc/unittest/__main__.py
@@ -16,5 +16,6 @@ from .thinking import *
 from .web_search import *
 from .models import *
 from .mcp import *
+from .tool_support_provider import *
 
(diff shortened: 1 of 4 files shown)


hlohaus:
@mvclaudianobj Are you familiar with the ToolSupportProvider? I find my current implementation to be more favorable. Could you please incorporate unit tests when integrating new features into core functionalities?

hlohaus:
@mvclaudianobj Thank you. Can you create a documentation on g4f.dev too?

PR #3353, opened 2026-02-08
fix: prioritize API key loaded from AuthManager over client-provided key

Duccioo:
This pull request improves authentication handling in the `OpenaiTemplate` provider by ensuring that API keys from the server environment are preferred over client-provided keys. This allows deployments (such as Dockerized environments) to securely use environment variables for authentication, even when the provider requires an API key to be set.

Authentication improvements:

* Updated `create_async_generator` in `OpenaiTemplate.py` to load the API key from the environment using `AuthManager` and prioritize it over any client-provided key. This ensures secure and consistent authentication, especially in server-side deployments.

### DIFF
--- a/g4f/Provider/template/OpenaiTemplate.py
+++ b/g4f/Provider/template/OpenaiTemplate.py
@@ -18,6 +18,7 @@ class OpenaiTemplate(AsyncGeneratorProvider, ProviderModelMixin, RaiseErrorMixin
     base_url = ""
     backup_url = None
     api_key = None
+    default_api_key = "secret"
     api_endpoint = None
     supports_message_history = True
     supports_system_message = True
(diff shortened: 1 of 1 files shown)


hlohaus:
@Duccioo, would it be possible to implement a configurable option for this feature? I require the ability to override API keys for my demonstration server. A straightforward approach would be to designate "secret" as the default API key, which would be overridden by any specified API key. Thank you for your consideration.

Duccioo:
I don't know if this is what you were looking for, but see if this feature can be useful.

bala-camp69:
Hey Duccioo, you can just set "secret" as the api_key to get the same results. And an env config like BrowserConfig.browser_port would be good. You can leave the PR open, it's good coding.

hlohaus:
@Duccioo Please add a config flag (env) for this feature.
```bash
set G4F_ENABLE_API_KEY_OVERRIDE=true
```

vyixor:
Well, I will say it's useful

hlohaus:
Hey @Duccioo , I've added the option to disable user API keys in gpt4free:

cls.disable_custom_api_key = os.environ.get("G4F_DISABLE_CUSTOM_API_KEY", str(cls.disable_custom_api_key)).lower() in ("true", "1", "yes")

It disables user API keys in some code lines.

PR #3367, opened 2026-02-25
fix: replace 34 bare except clauses with except Exception

haosenwang1018:
## What
Replace 34 bare `except:` clauses with `except Exception:`.

## Why
Bare `except:` catches `BaseException`, including `KeyboardInterrupt` and `SystemExit`, which can prevent clean process shutdown and mask critical errors. Using `except Exception:` catches all application-level errors while allowing system-level exceptions to propagate correctly.


hlohaus:
Thank you!

PR #3372, opened 2026-03-02
feat: implement secure filesystem tools for g4f models

imanpr33trai:
Adds a secure, audited filesystem tool suite allowing AI models to read, write, list, and search files within whitelisted directories. Includes path traversal protection, resource limits, and forbidden pattern blocking to ensure safe execution of file operations through tool calling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ lib64
 /.idea
 container/
 g4f.dev
+.env
--- a/g4f/providers/tool_support.py
+++ b/g4f/providers/tool_support.py
@@ -13,6 +13,15 @@ from .response import ToolCalls, FinishReason, Usage
 
 
 class ToolSupportProvider(AsyncGeneratorProvider):
+    """
+    Tool emulation provider for models that don't support native tool calling.
+
+    This provider adds system prompts to instruct models to return tool calls
+    in JSON format, then parses the response and converts it to OpenAI-compatible
+    tool call format.
+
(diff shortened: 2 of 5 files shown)


imanpr33trai:
This is my first contribution in my dev career, and I really wanted to contribute to this project.

Because I am making a small agent using gpt4free, and the models can't create, delete, edit, list, or access files properly, I added these features to help with those problems. If you think some more things should be added, please make sure to notify me.

I would really appreciate your feedback. If you think something needs to be improved or changed, please let me know. I’d be happy to work on it and continue contributing.

hlohaus:
Hey @imanpr33trai, thanks!

I've got a few questions:

Why aren't you using an MCP server? The tool calls in the body aren't OpenAI compatible.

Can it read/write all user files? Did you add verification for a specific directory?

Is it enabled for all requests, or can I enable it with a config? I don't want it used on my demo server for security reasons.

imanpr33trai:
_thank you for reviewing,_

i am now adding the mcp server to use the tool calls through it should i add the config.json type like configuration for the g4f about the permissions of the tool calls and human interaction before model using the tool 

### i am customizing or adding some new few things now like :

1.   MCP server replacing my hardcoded tools, so that every providers model can use it   **(REPLACING)**
2.   Configuration file (config.json) for customizing permission to which tool to use or not  **(NEW)**
3.   Human interaction  while in the request the model will ask the tool call to use or not  **(NEW)**
4.    some more security  **(IMPROVEMENT)**
 
**If you have any suggestions !  **
**PLEASE NOTIFY ME**

PR #3398, opened 2026-03-22
CI: Implement Python Matrix & Drop Legacy Support

abstract-333:
### Summary
Modernized the `Unittest` workflow by replacing sequential testing with a parallelized matrix strategy. This significantly reduces CI wait times and ensures compatibility across the latest Python releases.

### Key Improvements
* **Parallel Execution:** Added a matrix for **Python 3.10 through 3.14**, allowing tests to run concurrently.
* **Dropped Legacy 3.8:** Removed Python 3.8 support due to library incompatibilities.
* **Robustness:** Added `fail-fast: false` to ensure a single version failure doesn't kill the entire test suite.

### Verification:
* **Tested & Verified**: The updated workflow has been manually triggered and successfully passed across all matrix versions, [link](https://github.com/abstract-333/gpt4free/actions/runs/23406165073).

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -31,8 +31,6 @@ jobs:
     - name: Install requirements
       run: |
         pip install --upgrade pip
-        # Checking if min requirements exist before installing
-        if [ -f requirements-min.txt ]; then pip install -r requirements-min.txt; fi
         pip install -r requirements.txt
 
     - name: Run tests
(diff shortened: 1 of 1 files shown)


hlohaus:
@abstract-333 Thank you for contributing to gpt4free! However, this isn't quite the test matrix I'm looking for. The ideal flow should first check the minimum requirements and then move on to the full requirements. This approach works very well for this project.

abstract-333:
@hlohaus 
So, should we first test with the minimum requirements, then the **full requirements**, for every Python version?

hlohaus:
@abstract-333, you are able to verify the minimum requirements by exclusively considering the minimum Python version.

PR #3410, opened 2026-04-01
fix(OperaAria): real streaming, text+image coexistence, visible upload errors

yetval:
## What

Three bugs fixed in `g4f/Provider/OperaAria.py`:

**1. Streaming was not actually streaming**
Text chunks were appended to a buffer and yielded as one string after `[DONE]`. Now each chunk is yielded immediately as it arrives.

**2. Text was silently discarded when images were present**
The `elif text_buffer` check meant any text in a response containing images was thrown away. Text is now streamed inline and image responses are yielded independently after the loop.

**3. Media upload failures were swallowed silently**
`except Exception: continue` gave no indication when an image failed to upload. Now logs via `debug.log()` consistent with other providers.

## Why

- `supports_stream = True` was misleading — the provider was buffering the full response before yielding
- Mixed text+image responses lost all text content
- Silent upload failures made debugging impossible

### DIFF
--- a/g4f/Provider/OperaAria.py
+++ b/g4f/Provider/OperaAria.py
@@ -9,6 +9,7 @@ import base64
 import asyncio
 from aiohttp import ClientSession, FormData
 
+from .. import debug
 from ..typing import AsyncResult, Messages, MediaListType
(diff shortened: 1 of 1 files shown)


hlohaus:
Thank you

PR #3428, opened 2026-04-20
fix(OpenaiChat): use gizmo_interaction conversation_mode for custom GPTs

### LINKED ISSUE #3427: OpenaiChat: gizmo_id ignored — conversation_mode hardcoded to primary_assistant
## Description

When passing `gizmo_id` to `OpenaiChat` to use a custom GPT, the parameter is accepted but silently ignored. The HTTP payload always sends `conversation_mode: {"kind": "primary_assistant"}`, so the custom GPT's system instructions are never applied and the response comes from the base model instead.

## Root cause

In `g4f/Provider/needs_auth/OpenaiChat.py`, `create_authed` accepts `conversation_mode` as a parameter and correctly defaults it based on `gizmo_id` — but the two `data` dicts built inside the request loop hardcode the value instead of using the variable:

```python
# Line ~478 (api_key path)
"conversation_mode": {"kind": "primary_assistant"},  # should use conversation_mode

# Line ~539 (main path)
"conversation_mode": {"kind": "primary_assistant"},  # should use conversation_mode
```

## Fix

1. In `create_authed`, set `conversation_mode` based on `gizmo_id` when not explicitly provided:

```python
if conversation_mode is None:
    gizmo_id = kwargs.get("gizmo_id")
    if gizmo_id:
        conversation_mode = {"kind": "gizmo_interaction", "gizmo_id": gizmo_id}
    else:
        conversation_mode = {"kind": "primary_assistant"}
```

2. Replace the two hardcoded literals in the `data` dicts with the `conversation_mode` variable.

## Reproduction

```python
import g4f
from g4f.Provider import OpenaiChat

for chunk in g4f.ChatCompletion.create(
    model="gpt-4o",
    provider=OpenaiChat,
    messages=[{"role": "user", "content": "Hello"}],
    gizmo_id="g-XXXXXXXX",  # any custom GPT id
    stream=True,
):
    print(chunk)
# Response comes from base gpt-4o, not the custom GPT
```

## Version

g4f 7.5.0

vyguera:
## Summary

Fixes #3427

When `gizmo_id` is passed to `OpenaiChat`, the `conversation_mode` parameter in `create_authed` was never propagated to the HTTP request payload — both `data` dicts hardcoded `{"kind": "primary_assistant"}`, causing the custom GPT's system instructions to be ignored and the base model to respond instead.

## Changes

- In `create_authed`, when `conversation_mode` is `None` and `gizmo_id` is present in `kwargs`, default to `{"kind": "gizmo_interaction", "gizmo_id": gizmo_id}` instead of `primary_assistant`
- Replace the two hardcoded `conversation_mode` literals in the request payload `data` dicts with the `conversation_mode` variable

## Test

```python
import g4f
from g4f.Provider import OpenaiChat

for chunk in g4f.ChatCompletion.create(
    model="gpt-4o",
    provider=OpenaiChat,
    messages=[{"role": "user", "content": "Hello, follow your instructions"}],
    gizmo_id="g-XXXXXXXX",  # replace with a real custom GPT id
    stream=True,
):
    print(chunk, end="")
# Now responds using the custom GPT's system instructions
```

### DIFF
--- a/g4f/Provider/needs_auth/OpenaiChat.py
+++ b/g4f/Provider/needs_auth/OpenaiChat.py
@@ -451,7 +451,11 @@ class OpenaiChat(AsyncAuthedProvider, ProviderModelMixin):
                 conversation = copy(conversation)
 
             if conversation_mode is None:
-                conversation_mode = {"kind": "primary_assistant"}
+                _gizmo_id = kwargs.get("gizmo_id")
+                if _gizmo_id:
+                    conversation_mode = {"kind": "gizmo_interaction", "gizmo_id": _gizmo_id}
+                else:
+                    conversation_mode = {"kind": "primary_assistant"}
 
             if getattr(auth_result, "cookies", {}).get("oai-did") != getattr(conversation, "user_id", None):
                 conversation = Conversation(None, str(uuid.uuid4()))
@@ -475,7 +479,7 @@ class OpenaiChat(AsyncAuthedProvider, ProviderModelMixin):
                         "model": model,
                         "timezone_offset_min": -120,
                         "timezone": "Europe/Berlin",
-                        "conversation_mode": {"kind": "primary_assistant"},
+                        "conversation_mode": conversation_mode,
                         "system_hints": system_hints,
                         "supports_buffering": True,
                         "supported_encodings": ["v1"]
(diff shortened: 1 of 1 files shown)


hlohaus:
Thank you

PR #3429, opened 2026-04-21
Fix get_models ,Update Qwen provider models and thinking mode

3mora2:
Fix get_models
Refresh static model lists and improve dynamic model discovery/parsing. get_models now accepts an optional token, sends headers, and uses safe dict access (model.get(...)) to build text/image/vision/models lists and set default_model. Adjust generation logic so thinking_mode defaults to "Auto" when reasoning is enabled (otherwise "Fast"), introduce an auto_thinking variable, and propagate token headers into the StreamSession.


hlohaus:
Thank you

PR #3455, opened 2026-06-01
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade the MiniMax provider's model configuration to include the latest M3 model as the new default, and prune the legacy model list.

## Changes
- Add `MiniMax-M3` to the model selection list (placed first)
- Set `MiniMax-M3` as the new `default_model` and `default_vision_model`
- Retain `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as available alternatives
- Remove legacy models `MiniMax-Text-01` and `abab6.5s-chat` from the provider model list
- Update `model_aliases["MiniMax"]` to point at M3
- Add M3 entries to `g4f/providers/any_model_map.py` (forward + reverse mappings)
- Update unit tests to assert M3 is the default, M2.7 is retained, and the legacy models are removed

## Why
MiniMax-M3 is the new flagship model from MiniMax, offering a 512K context window, 128K max output tokens, and image input support across both the OpenAI-compatible and Anthropic-compatible interfaces. Adopting it as the default keeps users on the most capable model by default while leaving M2.7 and M2.7-highspeed available for callers that need the older behavior. Removing the long-deprecated `MiniMax-Text-01` and `abab6.5s-chat` IDs keeps the model list in line with the SKILL guidance to drop anything older than M2.7.

## Testing
- `python3 -m unittest etc.unittest.test_minimax -v` — all 9 tests pass
- Provider class loads cleanly and exposes the expected attributes
- `g4f/providers/any_model_map.py` imports without syntax errors

### DIFF
--- a/etc/unittest/test_minimax.py
+++ b/etc/unittest/test_minimax.py
@@ -19,9 +19,13 @@ class TestMiniMaxProvider(unittest.TestCase):
     def test_m27_highspeed_still_available(self):
         self.assertIn("MiniMax-M2.7-highspeed", MiniMax.models)
 
-    def test_old_models_still_available(self):
-        self.assertIn("MiniMax-Text-01", MiniMax.models)
-        self.assertIn("abab6.5s-chat", MiniMax.models)
+    def test_old_models_removed(self):
+        self.assertNotIn("MiniMax-Text-01", MiniMax.models)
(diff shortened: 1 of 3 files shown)


hlohaus:
@octo-patch thank you

PR #3465, opened 2026-06-13
[fix-aliyun_waf_aa]Update Qwen model lists and add request-id header

3mora2:
Refresh available Qwen model arrays (text_models, image_models, vision_models, models) to include newer variants (qwen3.7-*, qwen3.6-27b, qwen-latest-series-invite-beta-*) and remove/replace older entries. Change default_model to 'qwen3.7-plus'. Add an x-request-id header (UUID) to requests to mitigate Cloudflare aliyun_waf_aa errors.


hlohaus:
Thank you

PR #3468, opened 2026-06-22
🚀 Performance: Huge CLI & Backend Startup Optimization (87% Speedup) + Web UI Caching

kqlio67:
This PR drastically reduces the startup latency of the G4F framework (from ~3.1 seconds down to ~0.4 seconds) by eliminating eager imports across the provider and model registries. It also introduces aggressive static file caching for the Web UI.

## Key Changes
* **Lazy Provider Loading**: Introduced a PEP 562 dynamic dictionary in `g4f/Provider/__init__.py`. Providers are now mapped as strings and imported strictly on-demand.
* **Decoupled Model Definitions**: Replaced hardcoded class imports in `g4f/models.py` with string-based routing, preventing cascading library wakeups during basic `import g4f` calls.
* **CLI Routing Optimization**: Deferred heavy imports (like `gui.run`, `client`, and specific auth providers) inside `g4f/cli/__init__.py` so they are only loaded when their specific CLI command is executed.
* **Metadata Deferral**: Delayed the heavy `importlib.metadata` parse in `g4f/version.py` until the version is explicitly requested.
* **Static Asset Caching**: Added `max_age=31536000` (1 year) to `send_from_directory` responses in `g4f/gui/server/website.py` to leverage browser caching and significantly improve Web UI load times.
* **GUI API Stability**: Ensured backward compatibility in `gui/server/api.py` and `any_provider.py` by converting `Provider.__map__` into a smart dictionary (`_ConvertDict`) that auto-resolves strings back to classes seamlessly.

## Impact
Massive quality-of-life improvement for CLI users and developers embedding `g4f` into other Python environments where fast start/stop times are critical. Web UI users will also experience much faster subsequent page loads due to the new static caching. All unit tests pass seamlessly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ lib64
 /.idea
 container/
 g4f.dev
+har_and_cookies
--- a/g4f/Provider/OllamaSwarm.py
+++ b/g4f/Provider/OllamaSwarm.py
@@ -600,6 +600,9 @@ class OllamaSwarm(OpenaiTemplate):
         if not cls.models:
             cls.get_models()
 
+        if not model or model == "default":
+            model = getattr(cls, "default_model", "qwen3:14b")
+
         server_urls = cls.model_to_servers.get(model)
         if server_urls is None:
             resolved = cls.get_model(model)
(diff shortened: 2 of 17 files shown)


hlohaus:
Thank you👌

hlohaus:
@kqlio67 can you contact me on discord? The imports doesn't works anymore.

PR #3478, opened 2026-07-09
feat(providers): bypass Turnstile Captcha for DeepInfra using custom lightweight CDP client

kqlio67:
### What does this PR do?

**DeepInfra Provider - Turnstile Bypass**
Bypassing the Cloudflare Turnstile on the DeepInfra page requires blocking the original fetch request via JS and using very precise native CDP keyboard events (like simulating an OS-level 'Enter' key press) to get the token successfully. 
   
To achieve this highly reliable bypass, I wrote a custom, extremely lightweight CDP client (`Turnstile` in `g4f/Provider/deepinfra/turnstile.py`) that uses only the standard `websocket-client` library. It connects directly to Chrome, simulates human input flawlessly, and bypasses the captcha in just a few seconds without heavy dependencies. 
- Replaced the single `DeepInfra.py` file with a modular `deepinfra` directory.
- Uses the custom lightweight CDP client as the primary driver.
- Keeps `DrissionPage` as the reliable fallback option.

### Why is this needed?
- Improves the reliability of generating a captcha token for DeepInfra.
- Avoids forcing users to install heavy browser-automation libraries just to get a token.


hlohaus:
@kqlio67 great, i added only BrowserConfig.port support and fixed a error message

kqlio67:
> @kqlio67 great, i added only BrowserConfig.port support and fixed a error message

**Awesome, thanks for the review and the additions! Glad to help. Happy Community Day! 🎉**

hlohaus:
@kqlio67 You got a award:

kqlio67:
@hlohaus Wow, thank you so much! This truly means a lot to me 🙏 It's been a pleasure contributing to this project — the community and the codebase are both inspiring. Looking forward to contributing even more! 🚀

PR #3485, opened 2026-07-16
fix(CI): add timeout to get_models() requests to prevent CI hang

kqlio67:
## Problem

The `Unittest` CI workflow hangs indefinitely (2-6 hours) on `python -m etc.unittest` due to `requests.get()` calls without a `timeout` parameter in several provider `get_models()` methods.

The test passes `timeout=5` via `provider.get_models(timeout=5)`, but providers like `DeepInfra` and `glm` accept it as `**kwargs` and never use it — so the timeout is silently ignored.

When Cloudflare or other WAFs tarpit the connection (accept TCP but never respond), `requests.get()` waits indefinitely, blocking the entire test run.

### Provider timeout fix (7 files)
Added `timeout=kwargs.get("timeout", 15)` to `requests.get()` calls in `get_models()`:
- `DeepInfra.py`
- `glm/__init__.py`
- `local/Ollama.py`
- `needs_auth/Puter.py`
- `needs_auth/hf/HuggingChat.py`
- `needs_auth/hf/HuggingFaceInference.py` (also added missing `**kwargs` to method signature to prevent `NameError`)
- `needs_auth/hf/HuggingFaceMedia.py`

### CI safety net
Added `timeout-minutes: 5` to both `Run tests` steps in `unittest.yml` to prevent future hangs regardless of cause.

## Testing

Ran `python -m etc.unittest` on fork CI — all 167 tests passed in 19 seconds (28 skipped).

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -24,6 +24,7 @@ jobs:
       run: pip install -r requirements-min.txt
     - name: Run tests
       run: python -m etc.unittest
+      timeout-minutes: 5
     - name: Set up Python 3.14
       uses: actions/setup-python@v6
       with:
@@ -34,6 +35,7 @@ jobs:
         pip install -r requirements.txt
     - name: Run tests
       run: python -m etc.unittest
+      timeout-minutes: 5
     - name: Save PR number
       env:
         PR_NUMBER: ${{ github.event.number }}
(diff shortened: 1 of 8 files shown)


hlohaus:
Hey, you removed support for remote CDP hosts when BrowserConfig.port is set.
Can you fix that? @kqlio67

PR #3504, opened 2026-08-10
feat(provider): add OrcaRouter as a named provider

XiaoHuo888-hue:
## What

This adds a dedicated [OrcaRouter](https://www.orcarouter.ai) provider rather than relying on the generic OpenAI-compatible endpoint, mirroring how this repo already treats `OpenRouter`-style aggregators. Named routers such as `orcarouter/auto` pick an upstream per request. OrcaRouter is an OpenAI-compatible gateway that exposes 150+ models behind one API key, and it also provides gateway-level security controls for AI agents.

I'm an engineer on the OrcaRouter team.

## Changes

- **`g4f/Provider/needs_auth/OrcaRouter.py`** (new) — `OrcaRouter(OpenaiTemplate)` provider:
  - `base_url = "https://api.orcarouter.ai/v1"`, `default_model = "orcarouter/auto"`, `needs_auth = True`.
  - Overrides `get_headers` to inject `Authorization: Bearer {api_key}`, since the base `OpenaiTemplate` ships with that line commented out and without it requests return `401`.
- **`g4f/Provider/__init__.py`** — registers `OrcaRouter` in `_resolve_provider` (next to `OpenRouter`/`OpenRouterFree`) and in `_provider_names`, so it works with lazy lookup, `ProviderUtils.get_by_label`, and the API server.
- **`g4f/Provider/needs_auth/__init__.py`** — imports `OrcaRouter` alongside the other auth-gated providers.
- **`example.env`** — adds `ORCAROUTER_API_KEY=`. The key is auto-discovered via `AuthManager.load_api_key` (derives `ORCAROUTER_API_KEY` from the provider name), consistent with `OPENROUTER_API_KEY`.

## Why a named provider (not a generic OpenAI-compat passthrough)

gpt4free already models OpenRouter as a first-class named provider instead of pointing users at a bare `base_url` in the generic template. OrcaRouter is registered the same way, so the provider registry, model picker, and env-var handling stay consistent. Keys start with `sk-orca-`.

## Verification

- Unit tests (`etc.unittest.models`, `etc.unittest.config_provider`): 55 passed, 7 skipped — same before and after the change.
- Live L3 against the real gateway with a real key (via the registered provider path):
  - `get_models` → 192 models; catalog contains `orcarouter/auto`, `anthropic/claude-sonnet-4.6`, `deepseek/deepseek-chat`.
  - Chat `orcarouter/auto` → `PONG` (HTTP 200); chat `anthropic/claude-sonnet-4.6` → `PONG`.
  - Wrong key → `401 MissingAuthError` (fails loud).
  - Without the `get_headers` override the same chat call returns `401`, which is why the override is required.


hlohaus:
Thank you

PR #3507, opened 2026-08-26
Enhance DeepSeek streaming, uploads, and testing support

3mora2:
## Summary

This PR modernizes the authenticated DeepSeek provider to support the current web chat protocol, including reliable SSE streaming, proof-of-work challenges, file uploads, stream recovery, and comprehensive tests.

### Streaming and response recovery

- Added a dedicated SSE parser that preserves DeepSeek event types and supports snapshots, patches, batches, and close events.
- Added support for continuing `INCOMPLETE` responses and resuming interrupted streams.
- Tracks fragment types and indexes, including indexed `SET`, `APPEND`, `/type`, `/content`, and `-1` patches.
- Emits `THINK`, search, and tool fragments as reasoning while keeping `RESPONSE` and `TEMPLATE_RESPONSE` as visible output.
- Prevents duplicate content when a stream is resumed or replaced by a full-message snapshot.
- Attempts one bounded recovery when DeepSeek closes with `FINISHED` but provides no visible response.
- Raises a clear `ResponseError` if the recovered message still contains no response instead of silently returning an empty result.
- Never promotes hidden reasoning content to the final response.

### File uploads and proof of work

- Added endpoint-specific DeepSeek PoW challenge generation and solving.
- Runs the CPU-intensive PoW solver outside the async event loop.
- Added multipart file uploads with MIME type and filename handling.
- Supports multiple uploaded files while preserving their order.
- Waits for server-side file parsing and handles terminal parsing failures explicitly.
- Updated `detect_file_type` to consistently return the detected extension and MIME type.

### Provider structure and protocol handling

- Moved PoW and streaming logic into dedicated `deepseek` helper modules.
- Added normalized chat and upload headers.
- Added explicit completion payload construction and business-response envelope validation.
- Updated chat session creation, deletion, upload, completion, continue, and resume handling to match the observed DeepSeek web contract.
- Improved diagnostic logging without exposing response content or authentication data.

### Testing and diagnostics

- Added unit coverage for SSE parsing, fragment classification, status handling, resume and continue flows, empty responses, PoW, uploads, file parsing, and error responses.
- Added a reusable JSONL chunk logger for inspecting provider metadata and reasoning chunks during development.

## Why

DeepSeek streams responses as incremental fragment patches and can occasionally close a message as `FINISHED` after emitting reasoning without a final response. Previously, this could result in a successful request with an empty visible answer.

The updated implementation recovers the complete message when possible and returns an explicit error when DeepSeek genuinely provides no final response.

## Validation

- `python -m unittest etc.unittest.test_deepseek_stream etc.unittest.test_deepseek_upload`
  - 40 tests, OK
- `python -m etc.unittest`
  - 235 tests run
  - 9 skipped
  - Suite OK

Final verification was performed locally without issuing an additional live DeepSeek request.

### DIFF
--- a/etc/testing/deepseek_chunk_log.py
+++ b/etc/testing/deepseek_chunk_log.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+import json
+import uuid
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, TextIO
+
+from g4f.providers.response import JsonResponse, ProviderInfo, Reasoning
+
+
+def _json_default(value: Any) -> Any:
+    get_dict = getattr(value, "get_dict", None)
+    if callable(get_dict):
+        return get_dict()
+    if isinstance(value, Path):
+        return str(value)
+    if isinstance(value, (bytes, bytearray)):
+        return bytes(value).decode("utf-8", errors="replace")
+    attributes = getattr(value, "__dict__", None)
+    if isinstance(attributes, dict):
+        return {
+            key: item
+            for key, item in attributes.items()
+            if not key.startswith("__")
+        }
+    return str(value)
+
+
+class ChunkJsonlWriter:
+    """Append response metadata chunks to a durable, readable JSONL journal."""
+
+    def __init__(
+        self,
+        path: str | Path,
+        *,
+        run_id: str | None = None,
+    ) -> None:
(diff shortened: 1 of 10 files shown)


hlohaus:
Thank you @3mora2

